PR #300, opened 2022-12-15
Moved click handler for rename file to menuitem instead of button

sHtev:
keyboard actions were not invoking click handler, moving handler to `<MenuItem>` component fixes this


patch0 (approved):

IzzySmillie (approved):
🎉

PR #358, opened 2023-01-31
Small styling fixes

IzzySmillie:
- Fixes the scrollbar appearing in the visual output because of the border placement
- Stops the intermittent scrollbar flash on first load of app 
- Fixes the sense hat visual output height (was really small compared to designs)
- Fixes font size and icon visibility in the web component

-----
**BEFORE**
Screenshot 2023-01-31 at 14 04 15

**AFTER**
Screenshot 2023-01-31 at 14 44 43


src/components/AstroPiModel/AstroPiModel.scss
```diff
@@ -2,7 +2,6 @@
 @import '../../editor-spacing.scss';
 
 .sense-hat {
-  height: fit-content;
```
loiswells97, line 5: My only question on this would be, does it affect the web component in any way? I guess the reason it was so small (way too small as you say) was because we were trying to get the model + control panel to fit in the height. Just wondering whether this will change the way this is rendered for Astro Pi...?

IzzySmillie: It seems ok locally? 
Screenshot 2023-02-03 at 12 42 49

chrisruk (approved):

PR #362, opened 2023-02-03
Allowing HTML projects to render

IzzySmillie:
Part of: https://github.com/RaspberryPiFoundation/digital-editor/issues/64
If checking locally, checkout `html-starter-project` in the `editor-api` and run the `create_starter` rake task -> then go to `/projects/blank-html-starter` 

----
- Adds project type check and pushes `index.html` for html projects and `main.py` as default
- Adds tests for the requesting of a project for both project types

----
Screenshot 2023-02-03 at 14 52 28


loiswells97 (approved):
Looks great 👍  Looks like it's still connected up to the PythonRunner but I guess that's a problem for another day!

PR #376, opened 2023-02-20
Use new GraphQL API for all API requests on project index page

patch0:
## What's changed?

* Adds apollo-client, graphql packages
* Removes useProjectList hook]
* Fetches ProjectIndex data using GraphQL API, with fragments for pagination, project list item etc.
* Uses GraphQL API for DeleteProject and RenameProject modals

## Points for consideration

* When renaming a project the list doesn't get re-sorted by "updated at" time
* We're sorta changing state without telling Redux

## Before deployment

[editor-api#139](https://github.com/RaspberryPiFoundation/editor-api/pull/139) needs merging


IzzySmillie (approved):
LGTM! 

My only quarm is the pagination showing 1/1 when there's only one page but if we're changing to the [cursor](https://www.figma.com/file/2PjcTUsL1rHl1IaAht8yTm/Python-Editor---Dev-Handover?node-id=458%3A27640&t=hZnMHLRzyrSAzT6N-0) pagination in the future I think it's fine

PR #377, opened 2023-02-23
Remove reference to file index

loiswells97:
# What's Changed?

- Removed reference to file index to fix new project creation and saving a new file


IzzySmillie (approved):
Niiice. Looks good to me :shipit: 

There's probably some test re-jig we need to look into as this error didn't get caught! But another issue.

PR #378, opened 2023-02-23
Fix wrong runner bug

loiswells97:
# What's Changed?

- Tiny fix to use `HtmlRunner` for `html` projects so we can see what changes need to be made


IzzySmillie (approved):
What a nice simple fix! 🚀

PR #385, opened 2023-02-28
Tab discrepancy fix

IzzySmillie:
After increasing the hit area on the output toggle there is now a 2px difference between the heights of the tab container. Have added a `height: 50px` so that they stay the same height of `51px` (50px + 1px border)

**The bug:** 
See bug on staging: https://staging-editor.raspberrypi.org/ 
The text input is `49px` and the output is `51px`
Screenshot 2023-02-28 at 14 33 00
Screenshot 2023-02-28 at 14 33 04


loiswells97 (approved):
Looks good to me 👍

PR #395, opened 2023-03-02
fix codemirror hidden images

loiswells97:


patch0 (approved):
LGTM assuming code checks pass :+1:

PR #435, opened 2023-04-04
Copy fixes for log in / out

grega:
Fixes up: 

* login -> log in
* logout -> log out

to ensure consistency with how we refer to both elsewhere (and `log in / out` is the correct form for verbs).


grega:
Happy to add a changelog entry if needed, but seems a little unnecessary!

sra405 (approved):
👍

PR #457, opened 2023-04-24
Mission Zero tweaks

loiswells97:
# What's Changed

- Changed 3D model to re-render only on demand
- Refactored LED setting function to cache LED colour materials
- Refactored 3D model loading to use hooks and removed `three` symlink to `public` 
- Removed some unused libraries


sra405 (approved):
💪 great little refactor there and general tidy up! Noticed in my webpack bumps `three-stdlib` could be removed so good job!!

PR #483, opened 2023-04-28
feat(HtmlRunner): initial external link blocking

### LINKED ISSUE #469: Output panel links
Check the current behaviour for links in the output panel preview and determine best approach. Current suggestion is to block all links (which is believed to be the default behaviour) and invoke a modal when clicked to indicated it has been clicked but also why it's blocked.

Potentially use [this copy](https://www.figma.com/file/Rbn27JvH8A8wwqagKSReIq/Code-Editor---Component-library?type=design&node-id=1255-13480&t=Tcik0YFpxJ9vazpG-4) in a modal on click

sra405:
Closes #469


src/components/Editor/Runners/HtmlRunner/HtmlRunner.js
```diff
@@ -1,84 +1,147 @@
...
-    cssFiles.forEach((cssFile) => {
-      const cssFileBlob = getBlobURL(cssFile.content, "text/css");
-      indexPage = indexPage.replace(
-        `href="${cssFile.name}.css"`,
-        `href="${cssFileBlob}"`
+    const indexHTML = htmlFiles.find(
+      (component) => `${component.name}.${component.extension}` === "index.html"
+    );
+    const componentToPreview =
+      focussedComponent.extension === "html" ? focussedComponent : indexHTML;
```
danhalson, line 93: I might be missing something here, but is the TODO on line 88 still relevant now? it looks here like we're supporting any focussed file with the `*.html` extension

sra405: Ah ha! Good spot! That change was merged in from Lois's. I'll remove the TODO 👍

danhalson (approved):
Looks great, especially nice to see the PropTypes in here as I think it they'll benefit us across the editor-ui going forwards

PR #484, opened 2023-05-02
Remove unused package from addons

danhalson:
Removes the "@storybook/addon-links" from the addons list (since it's not defined in the package.json it won't work, and isn't used)


sra405 (approved):
👍

PR #446, opened 2023-04-13
Use proper logo, colors, and titles in Web App Manifest

cmfcmf:
The manifest was still referencing the defaults when this repo was created with create-react-app. For this to work, I also had to edit the logo to remove a superfluous white border and make it square.

I removed the `apple-touch-icon` for now, since it was also referring to the wrong logo. Eventually, it would probably make sense to properly generate this icon and other favicons using https://github.com/jantimon/favicons-webpack-plugin.

Ideally, the manifest would also contain PNG icons for older browsers that might not support SVGs, but I think using the SVG icon is a good first step.

PR #511, opened 2023-05-11
Use proper logo, colors, and titles in Web App Manifest (#446)

sra405:
From https://github.com/RaspberryPiFoundation/editor-ui/pull/446:

The manifest was still referencing the defaults when this repo was created with create-react-app. For this to work, I also had to edit the logo to remove a superfluous white border and make it square.

I removed the `apple-touch-icon` for now, since it was also referring to the wrong logo. Eventually, it would probably make sense to properly generate this icon and other favicons using
https://github.com/jantimon/favicons-webpack-plugin.

Ideally, the manifest would also contain PNG icons for older browsers that might not support SVGs, but I think using the SVG icon is a good first step.

---------


sra405:
No changes since the external PR approval so merging up after CI passed 👍

PR #527, opened 2023-05-26
Tweaks Projects site copy

grega:
Fixes some small inconsistencies with Projects site naming / references.


sra405 (approved):
👍 makes sense

Issue #551, opened 2023-06-29
Quick fix for Editor header on mobile


sra405:
Previously we had an issue with the panels going off the page when moving to a mobile devise but now the issue only remains for the header. A quick fix is needed as this will be actively changed and worked on soon as part of #539 BUT we want to go live with https://github.com/RaspberryPiFoundation/editor-ui/pull/540 this sprint.

https://github.com/RaspberryPiFoundation/editor-ui/assets/74183390/d48d4ee4-1061-4b3d-9b5a-161ce621dc18

PetarSimonovic:
Here are two ways of doing it so far by stacking the Download, Settings and Save buttons together. 

![Image](https://github.com/RaspberryPiFoundation/editor-ui/assets/69108995/dcbeec35-52a9-48e4-8214-2e6d0df2a896)




![Image](https://github.com/RaspberryPiFoundation/editor-ui/assets/69108995/0b8fb8f9-8821-4fb1-b3a4-eea0923eb536)

sra405:
Hey @PetarSimonovic really sorry it's taken this long!! The icons with text but move Downloads/Settings/Save below the icon option is best for now 👍  If you could just look at the spacing between the download and settings buttons if its a quick thing that'd be ace!

sra405:
closed with merge

PR #562, opened 2023-07-07
Issues/551 quick fix for editor header on mobile

PetarSimonovic:
Changes to css to address responsiveness issues with components in header. 

I've used a custom breakpoint to preserve the original style/layout as much as possible.

More details (and a video clip of the problem) are on the ticket:
[Quick fix for Editor header on mobile #551](https://github.com/RaspberryPiFoundation/editor-ui/issues/551)


danhalson (approved):
Looks great 👍

PR #580, opened 2023-07-12
Fix/text output overflow

sra405:
Resolves this support issue - https://raspberrypifoundation.slack.com/archives/C13FSSVDG/p1689004304702779


sra405:
dupe 🤦

Issue #554, opened 2023-06-29
Button


sra405:

PetarSimonovic:
Updated `design-system-react` with loaders and webpack config to use SCSS

[Pull request](design-system-react/pull/6)

sra405:
closed as part of https://github.com/RaspberryPiFoundation/editor-ui/pull/589

sra405:
splitting off remaining button uses into another backlog ticket. Aim of this ticket was to implement and theme for use in v2

PR #589, opened 2023-07-20
Issues/554 Button update

PetarSimonovic:
This PR is a first attempt at integrating the `design-system-core` Button with the `editor-ui`.

- the` design-system-react` button replaces the existing buttons in `LandingPage` and the Save button in `ProjectBar`. 
- the SCSS files for those components have changed to preserve existing styles and behaviour
- all overrides to the `rpf-button` styles from `design-system-core`  are now in `colours.scss` (for now - these may need to move into their own file to make overrides easier to maintain)

Details are in the [ticket](https://github.com/RaspberryPiFoundation/editor-ui/issues/554)


sra405 (approved):
👍 nice! It works nicely! I'm sure we'll change a few bits and pieces but it's in so glad we're using it and can iterate!

PR #604, opened 2023-07-25
Minor fixes (mostly buttons) to resolve issues in safari

sra405:
Minor updates to styles to fix:

- Button width
- Order on svg when selected

Both of which cause the editor to look like this in Safari

image

### DIFF
--- a/src/components/Editor/Runners/PythonRunner/OutputViewToggle.scss
+++ b/src/components/Editor/Runners/PythonRunner/OutputViewToggle.scss
@@ -3,9 +3,10 @@
 
 .output-view-toggle {
   display: inline-flex;
-  margin-right: $space-0-5;
+  vertical-align: middle;
+  margin-inline-end: $space-0-5;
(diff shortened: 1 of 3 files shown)


src/components/Editor/Runners/PythonRunner/OutputViewToggle.scss
```diff
@@ -3,8 +3,9 @@
 
 .output-view-toggle {
   display: inline-flex;
-  margin: auto $space-0-5 auto auto;
+  margin-right: $space-0-5;
```
magdalenajadach, line 6: I would use `margin-inline-end` as a logical property to make it rtl proof out of box, it will nicely flip sides when language is changed, [Margin examples](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values/Margins_borders_padding#margin_examples)

magdalenajadach (dismissed):
lgtm

PR #687, opened 2023-10-04
Style scaling to support integration into products with difffernt base font-size

sra405:
Due to integration of the web component into the project site a method of allowing editor styles to be utilised with a different root `font-size` was required as well as allowing the parent application to style aspects of the web component.

The README has been updated with the required info.

Initial integration into the project site example found here https://github.com/RaspberryPiFoundation/projects-ui/pull/2383


src/components/WebComponent/WebComponent.scss
```diff
@@ -2,11 +2,16 @@
 @use '../../font-size' as *;
 @use '../../colours' as *;
 
+:host {
+  font-size: 1.6rem;
+}
+
 #wc {
-  @include font-size-1(regular);
+  background: var(--editor-secondary-theme, transparent); // you can also access root variables here too
```
loiswells97, line 10: Can we switch this is a sass variable somehow?

src/font-size.scss
```diff
@@ -1,90 +1,90 @@
 @mixin font-size-5($line-height: regular) {
-  font-size: 5rem;
+  font-size: calc(5rem * var(--scale-factor, 1));
```
loiswells97, line 2: Where is `--scale-factor` being set and does it have to be a css, not a sass variable?

src/web-component.js
```diff
@@ -65,14 +65,18 @@ class WebComponent extends HTMLElement {
   }
 
   reactProps() {
-    return { ...this.componentAttributes, ...this.componentProperties };
+    return {
+      ...this.componentAttributes,
+      ...this.componentProperties,
+      forWebComponent: true,
```
loiswells97, line 71: What does this line do?

src/font-size.scss
sra405: it can be set in the parent app like [this](https://github.com/RaspberryPiFoundation/projects-ui/blob/e26e5d036db14ef97b8ca8096918b4dcb5e41a91/src/index.scss#L24) but if it's not set it'll default to 1. I'll play with sass variables but I'm not sure what'll take precedence when it's redefined 🤔

sra405: I can't seem to get the same default behaviour using sass variables. I either get the default 1.0 value used or a variable not defined error if I don't define it in that file 🤔 I've done a little searching too and annoying I can only find posts about theming with css rather than sass variables.
The [default value behaviour setup](https://sass-lang.com/documentation/variables/#default-values) _could_ work but it looks very setup for use within the same app and doesn't look like it would extend to web component well 😬

src/components/WebComponent/WebComponent.scss
sra405: This is the same as below and annoyingly I don't think so

PetarSimonovic (approved):
lgtm!

PR #709, opened 2023-10-18
Pass step to web component

loiswells97:


src/events/WebComponentCustomEvents.js
```diff
@@ -0,0 +1,17 @@
+const webComponentCustomEvent = (type, detail) =>
+  new CustomEvent(type, {
+    bubbles: true,
+    cancelable: false,
+    composed: true,
```
PetarSimonovic, line 5: Would we ever need to set other values for `bubbles`, `cancelable` or `composed`? Could we provide defaults but allow overrides? Or is that unnecessary?

PetarSimonovic (commented):
Just one query Lois

src/events/WebComponentCustomEvents.js
loiswells97: Not sure tbh, these are all we're using for now. I guess if we do need other values in the future we can generalise it a bit more?

PetarSimonovic (approved):
lgtm 👍

PR #712, opened 2023-10-19
Web component instructions

loiswells97:
# What's Changed?

- Added `instructions` attribute to web component and removed `step_number` attribute
- Added `InstructionsSlice` to contain data from the `instructions` attribute
- Converted web component attributes to camelCase in `web-component.js` so `WebComponentLoader` can have camelCase props

### DIFF
--- a/src/redux/InstructionsSlice.js
+++ b/src/redux/InstructionsSlice.js
@@ -1,0 +1,21 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+export const InstructionsSlice = createSlice({
+  name: "instructions",
+  initialState: {
+    steps: [],
+    stepNumber: null,
+  },
+  reducers: {
+    setStepNumber: (state, action) => {
+      state.stepNumber = action.payload;
+    },
+
+    setSteps: (state, action) => {
+      state.steps = action.payload;
+    },
+  },
+});
+
+export const { setStepNumber, setSteps } = InstructionsSlice.actions;
+export default InstructionsSlice.reducer;
--- a/src/redux/InstructionsSlice.js
+++ b/src/redux/InstructionsSlice.js
@@ -1,16 +1,10 @@
 import { createSlice } from "@reduxjs/toolkit";
+import { reducers } from "./reducers/instructionsReducers";
 
 export const InstructionsSlice = createSlice({
   name: "instructions",
   initialState: {},
-  reducers: {
-    setInstructions: (_state, action) => {
-      return action.payload;
-    },
-    setCurrentStepPosition: (state, action) => {
-      state.currentStepPosition = action.payload;
-    },
-  },
+  reducers,
 });
 
 export const { setCurrentStepPosition, setInstructions } =
--- a/src/redux/InstructionsSlice.js
+++ b/src/redux/InstructionsSlice.js
@@ -2,20 +2,17 @@ import { createSlice } from "@reduxjs/toolkit";
 
 export const InstructionsSlice = createSlice({
   name: "instructions",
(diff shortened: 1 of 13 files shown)


src/utils/camelCase.test.js
```diff
@@ -0,0 +1,5 @@
+import camelCase from "./camelCase";
+
+test("Converts string to camelCase", () => {
+  expect(camelCase("My_amazing_project-1")).toBe("myAmazingProject1");
+});
```
PetarSimonovic, line 5: This looks great Lois but it looks like we also use an npm package called [camelcase](https://www.npmjs.com/package/camelcase) that might do something similar

src/web-component.js
```diff
@@ -34,13 +35,22 @@ class WebComponent extends HTMLElement {
   }
 
   static get observedAttributes() {
-    return ["code", "sense_hat_always_enabled", "step_number"];
+    return ["code", "sense_hat_always_enabled", "step_number", "steps"];
```
PetarSimonovic, line 38: I wonder whether we need to decide as a team what we're passing through to the editor from projects, since I think [Magda's PR](https://github.com/RaspberryPiFoundation/projects-ui/pull/2392) might be giving the web component an entire project

src/web-component.js
```diff
@@ -34,13 +35,22 @@ class WebComponent extends HTMLElement {
...
+
+    if (["sense_hat_always_enabled"].includes(name)) {
+      value = Boolean(newVal);
+    } else if (["step_number"].includes(name)) {
+      value = parseInt(newVal);
+    } else if (["steps"].includes(name)) {
+      value = JSON.parse(newVal);
+    } else {
+      value = newVal;
+    }
```
PetarSimonovic, line 52: Just curious but is there a reason to check whether an array includes the name rather than `if name == "steps"`?

This could also be turned into a switch statement (although I appreciate that this might be a personal preference of mine!)

loiswells97: Yes, the main reason is so it can be extended when we add more attributes of different types

src/utils/camelCase.test.js
loiswells97: Ah this had completely passed me by 😅 Thanks, I've switched to use the library version instead 👍

src/redux/InstructionsSlice.js
```diff
@@ -0,0 +1,18 @@
...
+  name: "instructions",
+  initialState: {},
+  reducers: {
+    setInstructions: (_state, action) => {
+      return action.payload;
+    },
+    setCurrentStepPosition: (state, action) => {
+      state.currentStepPosition = action.payload;
+    },
+  },
```
sra405, line 13: 🔧 just being picky but it'd be good if this was setup initially so it was fully expandable and testable. I'd shift these reducers into the `./reducers` directory and pull together some quick initial tests. I imagine these are going to get more complicated

src/web-component.js
```diff
@@ -34,13 +35,20 @@ class WebComponent extends HTMLElement {
...
+    let value;
+
+    if (["sense_hat_always_enabled"].includes(name)) {
+      value = newVal === "true";
+    } else if (["instructions"].includes(name)) {
+      value = JSON.parse(newVal);
+    } else {
+      value = newVal;
+    }
+    this.componentAttributes[camelCase(name)] = value;
```
sra405, line 51: 🙌

loiswells97: 🎉

sra405 (approved):
👍 looks ace!! Think that refactor is definitely worthwhile

Issue #714, opened 2023-10-20
Update editor logout flow to destroy hydra session


sra405:
Following https://raspberrypifoundation.slack.com/archives/C042H46FHS7/p1696609100342819 the logout flow needs updating to remove the hydra session otherwise it'll remain and auto login when the login endpoint is hit.

Useful background:
- https://digital-docs.rpf-internal.org/docs/codebases/accounts/hydra-v1/logging-out
- https://raspberrypifoundation.slack.com/archives/G01QT0AF1LK/p1696839817712209

sra405:
various [methods exist in redux-oidc](https://github.com/IdentityModel/oidc-client-js/wiki#methods) one of which is `signoutRedirect` takes a `id_token_hint`, a generic `post_logout_redirect_uri` can be set in the user config and all is hooked up to the state so applies the ideal logout flow documented [here](https://digital-docs.rpf-internal.org/docs/codebases/accounts/hydra-v1/logging-out/#oidc-relay-party-rp-logout-flow) with minimal effort 💪

PR #715, opened 2023-10-20
Oidc-redux Signout Flow

sra405:
Related to #714 

Hits signoutRedirect with a `id_token` during logout to invalidate hydra session

### Note

This requires `post_logout_redirect_uris` to be registered with hydra client before merge (currently manual but I have an influx branch to bring this into terraform)


PetarSimonovic (approved):
lgtm! 👍

Issue #713, opened 2023-10-20
Allow links to our own domains


sra405:
Currently all links to external sources is blocked however some projects on the Projects Site step users through making links to rpf.io sources. Given we're in complete control of this domain all safeguarding issues should be fine (@MFarringtonRPF to confirm).

### Notes

It should be relatively straightforward to expand [this functionality](https://github.com/RaspberryPiFoundation/editor-ui/blob/main/src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx#L48C9-L48C21) with the one complexity being a partial match to the domain (ignoring paths).

Keep @pjbRPF informed

MFarringtonRPF:
If we can guarantee 100% control of this domain, agreed I think this is fine.

magdalenajadach:
PR: https://github.com/RaspberryPiFoundation/editor-ui/pull/729

PR #729, opened 2023-10-31
Allow links to our own domains

sra405:
Closing issue: https://github.com/RaspberryPiFoundation/editor-ui/issues/713

### DIFF
--- a/cypress/e2e/spec-html.cy.js
+++ b/cypress/e2e/spec-html.cy.js
@@ -43,34 +43,34 @@ it("updates the preview after a change when you click run", () => {
   getIframeBody().find("p").should("include.text", "hello world");
 });
 
-it("blocks external links", () => {
-  localStorage.clear();
-  cy.visit(baseUrl);
-  cy.get("div[class=cm-content]").invoke(
-    "text",
-    '<a href="https://raspberrypi.org/en/">some external link</a>',
-  );
-  cy.get(".btn--run").click();
-  getIframeBody().find("a").click();
-  cy.get("div[class=modal-content__header]")
-    .find("h2")
-    .should("include.text", "An error has occurred");
-});
+// it("blocks external links", () => {
+//   localStorage.clear();
+//   cy.visit(baseUrl);
+//   cy.get("div[class=cm-content]").invoke(
+//     "text",
+//     '<a href="https://raspberrypi.org/en/">some external link</a>',
+//   );
+//   cy.get(".btn--run").click();
+//   getIframeBody().find("a").click();
+//   cy.get("div[class=modal-content__header]")
+//     .find("h2")
+//     .should("include.text", "An error has occurred");
+// });
 
-it("allows internal links", () => {
-  localStorage.clear();
-  cy.visit(baseUrl);
-  cy.get("div[class=cm-content]").invoke("text", "<p>hello world</p>");
-  cy.get(".btn--run").click();
-  makeNewFile();
-  cy.get("div[class=cm-content]").invoke(
-    "text",
-    '<a href="index.html">some internal link</a>',
-  );
-  cy.get(".btn--run").click();
+// it("allows internal links", () => {
+//   localStorage.clear();
+//   cy.visit(baseUrl);
+//   cy.get("div[class=cm-content]").invoke("text", "<p>hello world</p>");
+//   cy.get(".btn--run").click();
+//   makeNewFile();
+//   cy.get("div[class=cm-content]").invoke(
+//     "text",
+//     '<a href="index.html">some internal link</a>',
+//   );
+//   cy.get(".btn--run").click();
 
-  const internalLink = getIframeBody().find("a");
-  internalLink.click();
-  const content = getIframeBody().find("p");
-  content.should("include.text", "hello world");
-});
+//   const internalLink = getIframeBody().find("a");
+//   internalLink.click();
+//   const content = getIframeBody().find("p");
+//   content.should("include.text", "hello world");
+// });
--- a/cypress/e2e/spec-html.cy.js
+++ b/cypress/e2e/spec-html.cy.js
@@ -1,4 +1,4 @@
-const baseUrl = "localhost:3010/en/projects/blank-html-starter";
+const baseUrl = "localhost:3000/en/projects/blank-html-starter";
 
 const getIframeDocument = () => {
   return cy
--- a/cypress/e2e/spec-html.cy.js
+++ b/cypress/e2e/spec-html.cy.js
@@ -1,4 +1,4 @@
-const baseUrl = "localhost:3000/en/projects/blank-html-starter";
+const baseUrl = "http://localhost:3000/en/projects/blank-html-starter";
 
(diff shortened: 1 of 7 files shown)


cypress/e2e/spec-html.cy.js
```diff
@@ -57,6 +57,17 @@ it("blocks external links", () => {
...
+  localStorage.clear();
+  cy.visit(baseUrl);
+  cy.get("div[class=cm-content]").invoke(
+    "text",
+    '<a href="https://www.rpf.io/seefood">some external link</a>',
+  );
+  cy.get(".btn--run").click();
+  cy.url().should("not.equal", baseUrl);
+});
+
```
loiswells97, line 70: I think I don't quite understand this test... Because we don't seem to be simulating the click on the actual link (that I can see). Although I'm not sure how easy it would be to test the content of the external link page, especially if we just put in the preview iframe as Marc was suggesting 🤔  Unless what we want to do is run the code, find the link in the iframe and then check that the href has not been replaced...

src/components/Editor/Runners/HtmlRunner/HtmlRunner.test.js
```diff
@@ -237,11 +244,11 @@ describe("When run is triggered", () => {
...
 });
 
-describe("When an external link is rendered", () => {
+describe("When a non-permitted external link is rendered", () => {
   let store;
   const input =
-    '<head></head><body><a href="https://google.com">EXTERNAL LINK!</a></body>';
-  const output = `<head></head><body><a href="javascript:void(0)" onclick="window.parent.postMessage({msg: 'ERROR: External link'})">EXTERNAL LINK!</a><meta filename="index.html" ></body>`;
+    '<head></head><body><a href="https://google.test/">EXTERNAL LINK!</a></body>';
+  const output = `<head></head><body><a href=\"javascript:void(0)\" onclick=\"window.parent.postMessage({msg: 'ERROR: External link'})\">EXTERNAL LINK!</a><meta filename=\"index.html\" ></body>`;
```
loiswells97, line 251: We've got some linting warnings on this

cypress/e2e/spec-html.cy.js
PetarSimonovic: Yes you're right about this! My error

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -104,21 +113,41 @@ function HtmlRunner() {
...
+      if (linkElement) {
+        linkElement.addEventListener("click", (e) => {
+          e.preventDefault();
+          const isExternalLink = matchingRegexes(
+            allowedExternalHrefs,
+            linkElement.href,
+          );
+          const target = isExternalLink ? "_blank" : "_top";
+          window.open(linkElement.getAttribute("href"), target);
+        });
```
PetarSimonovic, line 149: This needs to change in light of Mark's comments

loiswells97: 👍

cypress/e2e/spec-html.cy.js
```diff
@@ -57,6 +57,20 @@ it("blocks external links", () => {
...
+  cy.get("div[class=cm-content]").invoke(
+    "text",
+    '<a href="https://rpf.io/seefood">some external link</a>',
+  );
+  cy.get(".btn--run").click();
+  const externalLink = getIframeBody().find("a");
+  externalLink.click();
+  cy.url().should("be.equals", baseUrl);
+  cy.get("div[class=modal-content__header]").should("not.exist");
+});
```
PetarSimonovic, line 72: Cypress test now checks that the url hasn't changed (so we're definitely on the same page after the link is clicked) and that no modal appears. But: 
- Should we check to make sure it's not an error modal?
- Should we check that the iframe has some content that will confirm it's loaded the page? I'm reluctant to check for specific content since if the `seefood` project changes then that will break this test. (I don't think it's an issue if the `seefood` project is taken down completely since the link will still be permitted and will just render a missing page)

loiswells97: I think what you've got it fine 👍

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -104,21 +113,47 @@ function HtmlRunner() {
       if (typeof event.data?.msg === "string") {
         if (event.data?.msg === "ERROR: External link") {
           setError("externalLink");
-        } else {
+        } else if (event.data?.msg !== "Allowed external link") {
```
loiswells97, line 116: Can we switch these around to make it clearer? I.e. can we have `if (event.data?.msg === "Allowed external link") <do allowed external link stuff> else <render html from user's code>`

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -104,21 +113,47 @@ function HtmlRunner() {
...
   };
 
   const iframeReload = () => {
+    console.log("IFRAME RELOADED");
     const iframe = output.current.contentDocument;
     const filename = iframe.querySelectorAll("meta[filename]")[0]
       ? iframe.querySelectorAll("meta[filename]")[0].getAttribute("filename")
       : null;
     if (runningFile !== filename) {
+      console.log("Setting filename");
```
loiswells97, line 134: There's a number of console logs and a commented line that could do with going

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -83,6 +92,18 @@ function HtmlRunner() {
...
+    let filename;
+    if (iframe) {
+      filename = iframe.querySelectorAll("meta[filename]")[0]
+        ? iframe.querySelectorAll("meta[filename]")[0].getAttribute("filename")
+        : externalLink;
+    } else {
+      filename = externalLink;
+    }
+    return filename;
+  };
```
PetarSimonovic, line 105: Would prob be better to return from the `if` then ditch the `else` statement I think

loiswells97: I don't mind - I tend to find things that return early a bit confusing, but that's just personal preference

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -168,70 +209,88 @@ function HtmlRunner() {
   }, [runningFile]);
 
   const runCode = () => {
-    setRunningFile(previewFile);
```
loiswells97, line 171: I don't really understand why this was deleted?

PetarSimonovic: Hmmm must be an error on my part. I've reinstated

PetarSimonovic (approved):
lgtm! 👍

Issue #733, opened 2023-11-03
Un-spaghetti the HtmlRunner 


magdalenajadach:
The logic around internal/external links is very nested and complicated, it would be nice to take it out to a separate file

Related to: https://github.com/RaspberryPiFoundation/editor-ui/issues/713

PetarSimonovic:
### Tests
Tests need to be updated to capture the current behaviour. 
- [ ] When an external link is loaded, the output tab contains the correct name of the page.
- [ ] After a user has navigated away from an external link, the external link is removed from state (this is happening now, but previously the link persisted in state, which meant that clicking on any other link (external or not) would navigate to the external link again)

sra405:
We need to be able to test what's rendered in the iframe. Ideally this would happen before the render as diving into an iframe with jest is awkward. Consider pulling out `indexPage` and/or `runCode` to make testing the iframes html much easier.

PR #748, opened 2023-11-10
Initial javascript support in html projects

sra405:
Small tweaks to allow javascript files in HTML projects rather than relying on inline styles


PetarSimonovic (commented):
This all works great Scott! Do we need any tests since HTMLRunner is going to be worked on and untangled soon?

sra405:
> This all works great Scott! Do we need any tests since HTMLRunner is going to be worked on and untangled soon?

Ideally yes however I really can't see a way of testing the code in the HTML iframe in it's current state 🤦 so I've added a [little note here](https://github.com/RaspberryPiFoundation/editor-ui/issues/733#issuecomment-1822274838) as a reminder because we'll likely want a bunch of tests to check this logic!

PetarSimonovic (approved):
lgtm! 👍

PR #752, opened 2023-11-14
Allow themes to be set and toggled globally

sra405:
Adds a toolbar item that then wraps the story in the classes needed to theme the editor in light or dark mode.

https://github.com/RaspberryPiFoundation/editor-ui/assets/74183390/0d825ee6-27bd-45e2-a3b9-c209be25a8de

### DIFF
--- a/storybook/.storybook/preview.js
+++ b/storybook/.storybook/preview.js
@@ -7,14 +7,15 @@ import { withRouter } from "storybook-addon-react-router-v6";
 
 import i18n from "utils/i18n"; // importing the same i18n
 
+import "assets/stylesheets/App.scss";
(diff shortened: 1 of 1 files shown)


storybook/.storybook/preview.js
```diff
@@ -60,3 +63,17 @@ export const parameters = {
     states: [],
   },
 };
+
+export const globalTypes = {
```
PetarSimonovic, line 67: It's great to know about this. Seems very straightforward and could be really useful in the future for implementing all sorts of custom controls (eg, I think Storybook has it's own responsiveness sizes that might not align with ours)

PetarSimonovic (approved):
lgtm! 👍

Issue #745, opened 2023-11-10
Login to the host app from web-component


sra405:
We need the functionality to trigger login and signup via the host application from the web component (i.e. in the [new downloads sidebar panel](https://www.figma.com/file/SrINjD1MCT6VWYf1Yc9fV1/Code-Editor---Designs?type=design&node-id=3734-69935&mode=design&t=2gODKub9tIgfQkt1-4)).

Initial thinking:
- fire off login or signup custom events from the web component
- pick these events up by the host
- trigger the hosts typical login/signup flow

PetarSimonovic:
- [ ] Users can log in or sign up to the projects site using these buttons~ (https://github.com/RaspberryPiFoundation/editor-ui/issues/745)
  - [ ] They are then also logged into to Editor 
  - [ ] When logged in, their progress within the Project and their accompanying Editor code is saved, so that they can return to it

PetarSimonovic:
Ticket [#740](https://github.com/RaspberryPiFoundation/editor-ui/issues/740) offers further context on the log in, sign up and download functionality

PR #767, opened 2023-11-17
Issues/745 login button custom event

PetarSimonovic:
-  `DownloadPanel` adds custom events for its Log In and Sign Up buttons to the DOM.
-  The Projects site will have corresponding event listeners that will detect when these events are fired

Part of #745


src/events/WebComponentCustomEvents.js
```diff
@@ -15,3 +15,9 @@ export const runStartedEvent = webComponentCustomEvent("editor-runStarted");
 
 export const stepChangedEvent = (detail) =>
   webComponentCustomEvent("editor-stepChanged", detail);
+
+export const logInEvent = (detail) =>
+  webComponentCustomEvent("editor-logIn", detail);
+
+export const signUpEvent = (detail) =>
+  webComponentCustomEvent("editor-signUp", detail);
```
loiswells97, line 23: You only need to add the detail if we're passing extra data to the event (see the `codeChangedEvent` for instance), but also if you want to keep it like this for consistency I don't mind...

src/components/Menus/Sidebar/DownloadPanel/DownloadPanel.test.js
```diff
@@ -76,6 +85,22 @@ describe("DownloadPanel", () => {
...
+  });
+
+  test("the sign-up button calls a signUpHandler when clicked", async () => {
+    const signUpButton = screen.getByText(
+      "downloadPanel.signUpButton",
+    ).parentElement;
+    fireEvent.click(signUpButton);
+    await waitFor(() => expect(signUpHandler).toHaveBeenCalled());
+  });
+
```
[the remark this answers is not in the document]
PetarSimonovic: Not needed - removed!

src/events/WebComponentCustomEvents.js
PetarSimonovic: Removed!

loiswells97: On this, I guess there are two options for consistency here: either make the events with no detail all events or all functions that just return the event... 🤔 Either way I guess we'll need to change how some of them are referenced (either `myCustomEvent` or `myCustomEvent` depending on which we go with)

loiswells97 (approved):
Looks great 👍

PR #772, opened 2023-11-20
Issues/759 download panel hide login prompts when logged in

PetarSimonovic:
DownloadPanel doesn't display log in prompts or buttons if the user is already logged in


src/components/Menus/Sidebar/DownloadPanel/DownloadPanel.jsx
```diff
@@ -12,6 +13,7 @@ import "../../../../assets/stylesheets/DownloadPanel.scss";
 
 export const DownloadPanel = () => {
   const { t } = useTranslation();
+  const user = useSelector((state) => state?.auth?.user);
```
loiswells97, line 16: We've never needed the `?`s before 🤔 Wondering whether this is coming from the tests, and if so I think we've been representing a non-logged in user with `auth: {}` (so this shouldn't fail if there's no user)

src/components/Menus/Sidebar/DownloadPanel/DownloadPanel.test.js
```diff
@@ -4,6 +4,7 @@ import { Provider } from "react-redux";
 import { MemoryRouter } from "react-router";
 import configureStore from "redux-mock-store";
 import FileSaver from "file-saver";
+import { error } from "console";
```
loiswells97, line 7: There's a lint error here ☝️

PetarSimonovic: Removed - thanks!

src/components/Menus/Sidebar/DownloadPanel/DownloadPanel.jsx
PetarSimonovic: We use optional chaining in the `ProjectsPanel` and the `ProjectsIndex` but I've removed it here for now and passed in an empty `auth` for the tests.

Is there a reason not to use optional chaining? Could it be a useful guard just in case the auth or the user is missing?

Anyway, that might be a discussion for another time so I've removed for now!

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Update design-system-react dependency to publicly available version
 - Storybook SVG loader added and store configured (#749)
 - Switched `FontSizeSelector` and `ThemeToggle` over to using the new `SelectButtons` (#757)
+- Download panel removes log-in prompts for logged-in users
```
loiswells97, line 31: Can we have the PR number in here?

loiswells97 (approved):
Looks great, let's get it in!

Issue #708, opened 2023-10-17
WebComponent `#root` doesn't have a height attribute


patch0:
I've updated the Web Component to v0.19.2 on [a review app](https://astro-pi-reg-pr-1207.herokuapp.com/mz/code_submissions/new), and there are a couple of niggles, that I think are probably related.

* The web component doesn't fill the container vertically (`height: 100%`) has been removed off the `#root` element, and I'm not sure how to change this from outside the shadow dom.
* When you add more lines, it overflows the containing element, and makes the rendered pi massive :) :pie: 
* Dark mode doesn't have a dark "outside", i.e. inside the containing element of the page.

Compare with the [current release](https://staging-missions.astro-pi.org/mz/code_submissions/new)

sra405:
also consider how we remove the scrollbar for the web component too, this isn't cutting it
```
::-webkit-scrollbar {
    display: none;
}
```

loiswells97:
Could be related? Need to fix weird width issues with the web component e.g. where the web component ends up halfway between desktop and tablet modes or extends beyond the page width

loiswells97:
### Findings

- You can style the editor web component root from the host page using `::part(editor-root)`
- Putting `::part(editor-root) { height: 100%; }` fixes the height issues mentioned, and the background can be set inside this too
- It looks like the background is intentionally left white so we can theme the web component in the projects site
- The question remains whether we can apply these styles to the editor root in the editor itself, or whether it's best for host pages to style the web component root themselves

sra405:
Just noticed a minor media/container query change that's currently breaking input/output panels 😄 the rest should be good and configurable in the host page like you've detailed @loiswells97 👍 
https://github.com/RaspberryPiFoundation/editor-ui/pull/776

PR #776, opened 2023-11-22
set breakpoints on container not media query

sra405:
Related to https://github.com/RaspberryPiFoundation/editor-ui/issues/708 and fixes input/output panel breakpoints when web component is positioned on a page that doesn't take the full width of the screen

Currently the breakpoints change as a result of the positioning of the web component on the page e.g. if additional margin is added around the web component the breakpoints will be effected

https://github.com/RaspberryPiFoundation/editor-ui/assets/74183390/371ce394-6fb8-4e97-bb67-2e32e58ef4b9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


loiswells97 (approved):
Looks great 👍

PR #793, opened 2023-11-27
DownloadIcon wrapper replaced with svg

PetarSimonovic:
- DownloadIcon wrapper replaced with correct svg

For context, Max's explanation is:
- the original download.svg was exported without the whitespace so it was a 16x16px rather than a 24x24px
- without the white space all the icons will be different sizes
- we will probably find other icons that have previously been exported incorrectly and will need to be replaced


loiswells97 (commented):
Looks great overall! My only comment is a very nit-picky one, but it looks like the icon is no longer centred vertically on the download button
Screenshot 2023-11-27 at 12 25 24

loiswells97 (approved):
Looks good to me 👍

PR #807, opened 2023-12-05
Instructions styling fixes

loiswells97:
Fix gap between different instruction blocks
Fix multi-line output block


sra405 (approved):
👍 lgtm

PR #820, opened 2023-12-12
RPi Org link opens in new tab.

### LINKED ISSUE #830: RPi link should open in a new tab
When clicking on the RPi logo and text link at the top left should open in a new tab rather then taking the user away from their code.

ownimage:
This change makes the RPi Org link at the top left opens in new tab.

closes #830


sra405 (approved):
👍 sorry this has taken so long for us to get to, thanks for your contribution 😄

PR #828, opened 2023-12-15
Issues/812 quiz custom event

PetarSimonovic:
- adds a custom event to notify Projects when the quiz is loaded


loiswells97 (approved):
This looks good to me 👍

PR #852, opened 2023-12-19
Extracting user from localStorage

### LINKED ISSUE #743: Get web component auth auto-refresh working
### Potential Approach

One way of achieving this may be to pass the user object from the host page's store into the web component as an attribute instead of retrieving from `localStorage`. This way the web component should re-render when the user's credentials change. However, we would have to consider whether this approach is secure enough.

sra405:
closes #743

The idea is to intercept editor actions (i.e. actions that could call the API) and update the user object in state prior to that call being made. That way if the user object changes for whatever reason (like cred renewal) the latest user object will be used for the API call


src/containers/WebComponentLoader.test.js
```diff
@@ -147,7 +155,7 @@ describe("When loaded as a cold user", () => {
 
     test("Calls useProjectPersistence hook with correct attributes", () => {
       expect(useProjectPersistence).toHaveBeenCalledWith({
-        user,
+        user: undefined,
         project: { components: [] },
         hasShownSavePrompt: false,
         justLoaded: false,
```
loiswells97, line 161: I'm a bit confused by this bit - why are we expecting the user to be undefined even though they're logged in? Is it because the user only gets set when the user attempts to call an action, but not initially? 🤔

sra405: Yea its exactly that. So all we can really test is calls are made to the dispatch functions with the inputs we expect, they don't actually change state. So this will always be `undefined` (notice it's not `null` which it would be off that initial `setUser` action). So I've added the tests to check for the `setUser` dispatches here too

src/redux/middlewares/localStorageUserMiddleware.test.js
```diff
@@ -0,0 +1,80 @@
...
+      };
+      store = mockStore(initialState);
+      store.dispatch = jest.fn();
+    });
+
+    describe(`when authKey has been updated`, () => {
+      beforeEach(() => {
+        localStorage.setItem("authKey", "some-key");
+        localStorage.setItem("some-key", JSON.stringify(updatedUser));
+      });
```
loiswells97, line 70: Not sure I understand in what circumstances the auth key would ever be updated - I think it's pretty standard if I'm not wrong 🤔

loiswells97 (commented):
Just a couple of questions I'm not sure about in the tests and a missing changelog entry, otherwise looking good

src/redux/middlewares/localStorageUserMiddleware.test.js
sra405: yea sorry, badly named. This is testing when the authKey value has changed so in this case "some-key" 😅

loiswells97: Ohhhh, okay, so you mean the local storage user has been updated rather than the auth key?

sra405: yes! That's 100% a better way of putting it 😅 I'll update that!

loiswells97 (approved):
Yes, let's get this in!

PR #864, opened 2024-01-02
Fix project height for standalone editor

sra405:
Currently due to the styling changes required to maximise the height of the web component in Firefox the standalone editor took up too much space (didn't account for the top nav). This fix applies the required `min-block-size` only to the web component


magdalenajadach (approved):
lgtm

PR #865, opened 2024-01-04
Fix width of new file button

loiswells97:
Before:
Screenshot 2024-01-04 at 15 41 01

After:
Screenshot 2024-01-04 at 15 41 23


sra405 (approved):
👍 nice

PR #881, opened 2024-01-11
Fix web-component boolean attributes

sra405:
Fix so anything other than explicitly setting the boolean attributes as false is taken as true

### DIFF
--- a/src/web-component.js
+++ b/src/web-component.js
@@ -58,7 +58,7 @@ class WebComponent extends HTMLElement {
         name,
       )
     ) {
-      value = newVal === "true";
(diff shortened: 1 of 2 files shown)


src/web-component.js
```diff
@@ -106,6 +106,8 @@ class WebComponent extends HTMLElement {
       this.root = ReactDOMClient.createRoot(this.mountPoint);
     }
 
+    console.log(this.reactProps());
```
magdalenajadach, line 109: could you please remove the 'console.log' bit

magdalenajadach (approved):
lgtm

PR #885, opened 2024-01-12
Web component on port 3011

sra405:
Move fully to port 3011 due to web component builds being reliant on it locally and docker port mapping not cutting the mustard


sra405:
completed as part of another PR and live

PR #893, opened 2024-01-18
Add a Pyodide runner based on the python-execution-prototypes

tuzz:
This pull request adds experimental support for running Python programs using Pyodide. The code is based on the [python-execution-prototypes](https://github.com/RaspberryPiFoundation/python-execution-prototypes) and the [technical analysis](https://docs.google.com/document/d/1mpKliJaqYYkPJR1hGLZxUDhNFOwgBm69nc70iiVj_3I/edit#heading=h.3orc9tdzyy4b) document. The related GitHub issue is https://github.com/RaspberryPiFoundation/editor-ui/issues/891.

The PyodideRunner.jsx can be used instead of the current PythonRunner.jsx (Skulpt) by provided a `?pyodide=true` parameter in the URL as shown in the screenshot below:

Screenshot 2024-01-30 at 11 39 01

### Notes

The PyodideRunner supports the following features:
- Execution of Python programs
- Importing of [most of](https://pyodide.org/en/stable/usage/wasm-constraints.html) Python's standard library
- Importing of [many common](https://pyodide.org/en/stable/usage/packages-in-pyodide.html) Python packages
- Importing of pure Python packages from PyPi
- Stopping execution
- Sending standard input to Python programs
- Closing the standard input stream by pressing `Ctrl-d`
- Visualising turtle graphics using the [modified turtle package](https://github.com/RaspberryPiFoundation/turtle)
- Visualising pygal charts using the [modified pygal package](https://github.com/RaspberryPiFoundation/pygal.js)

Due to time constraints, the following tasks remain outstanding:

- Visualising sense_hat graphics using the [modified sense_hat package](https://github.com/RaspberryPiFoundation/sense_hat)
  - There is a work-in-progress pull request [here](https://github.com/RaspberryPiFoundation/editor-ui/pull/913)
- Apply HTTP headers in a service worker (see https://github.com/RaspberryPiFoundation/editor-ui/issues/903)
  - Note that `crossorigin` will need to be added to cross-domain resources once the HTTP headers are added
- Add unit/integration tests that run real Python programs through the PyodideRunner.jsx
- Manual testing for backwards compatibility of existing Python programs, projects and the WebComponent
- Implementing the p5 package to work with Pyodide
  - See https://github.com/RaspberryPiFoundation/editor-ui/issues/949
- Implementing a dedicated SQL editor for use in the Ada CS project
  - Note that Pyodide supports importing `sqlite3` so we can add a UI wrapper around it

PR #895, opened 2024-01-18
Update yarn from version 1.22.21 to 3.4.1 and webpack-dev-server from 3.11.1 to 4.0.0

tuzz:
We need to be able to conditionally set headers for pyodideWorker.js in pull request https://github.com/RaspberryPiFoundation/editor-ui/pull/893 but this feature isn’t supported in webpack-dev-server 3.x.x. We therefore need to upgrade to version 4.0.0 which allows a function to be provided in the headers option. When attempting this upgrade, I ran into [this issue](https://github.com/storybookjs/storybook/issues/22431) and only managed to resolve it by also updating yarn to version 3.

I’ve tried to follow the migration guide here:
https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Minor copy changes to HTML add file modal
 - Toggle errors sent via apiCallHandler off (#890)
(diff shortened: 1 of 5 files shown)


magdalenajadach (approved):
lgtm

PR #901, opened 2024-01-23
Code and assets identifiers

### LINKED ISSUE #896: Split identifiers into asset and code identifiers
### 2. Split identifiers into asset and code identifiers
Assets are loaded by:
* asset identifier

Code is loaded by:
* code identifier, OR
* `code` prop, ELSE
* blank

This has been done by:
1. Update `editor-ui` to allow two identifiers: `asset-identifier` and `code-identifier`
2. Update `editor-api` to add endpoint for assets only (easy, as above)
3. Update `editor-api` to add endpoint for code only (easy)

#### Draft spike
This has been drafted out in [the associated branch](https://github.com/RaspberryPiFoundation/editor-ui/pull/888) which splits out the asset loading from the project loading. 

Using Redux devtools we can see the assets from the specified project are available to the user.

For HTML projects, as we can't provide the project type via props, this can be achieved loading a blank HTML project which will in turn set the project type.

You can preview this using these project identifiers:
```
  <editor-wc
    code_identifier="top-5-emoji-list-step-5" 
    class="c-editor__wc" 
    asset_identifier="anime-expressions"
    output_only
    data-editor-target="editor"
  >
```
#### Pros
* A comprehensive, robust, reliable option as it allows specific loading of different assets and code, without having to overwrite code. This should prevent mistakes or unstable situations.
* API work required is simple, just splitting the response into two endpoints, one for the assets and the other for the code
* Backwards compatible, the new assets and code props will supplement the existing `identifier` prop
  
#### Considerations
* There will be an increase in API calls, as a project with separate identifiers will now be making two calls.
* This is mitigated by preserving existing `identifier` prop and routes which will load both assets and code when they're the same, with the new props only being used when the code and assets are different.

_Originally posted by @conorriches in https://github.com/RaspberryPiFoundation/editor-ui/issues/868#issuecomment-1895657837_
 
## Notes

- this is work for the block-to-text alpha so only needs to be merged into the https://github.com/RaspberryPiFoundation/editor-ui/tree/output-only-wc feature branch
- any additional api endpoints should be fine going into production but sanity check first
- enables assets and code to be split and managed in the backend 🥳 
- current assumption is we'll hardcode the code template in app and pass populated code back as `code`

loiswells97:
closes #896

### DIFF
--- a/src/hooks/useProject.js
+++ b/src/hooks/useProject.js
@@ -10,6 +10,8 @@ export const useProject = ({
   code = null,
   accessToken = null,
 }) => {
+  const project = useSelector((state) => state.editor.project);
+  const loading = useSelector((state) => state.editor.loading);
   const isEmbedded = useSelector((state) => state.editor.isEmbedded);
   const isBrowserPreview = useSelector((state) => state.editor.browserPreview);
   const getCachedProject = (id) =>
@@ -66,4 +68,35 @@ export const useProject = ({
     const data = defaultPythonProject;
     dispatch(setProject(data));
   }, [projectIdentifier, cachedProject, i18n.language, accessToken]);
+
+  useEffect(() => {
+    if (code && loading === "success") {
+      const defaultName = project.project_type === "html" ? "index" : "main";
+      const defaultExtension = project.project_type === "html" ? "html" : "py";
+
+      const mainComponent = project.components?.find(
+        (component) =>
+          component.name === defaultName &&
+          component.extension === defaultExtension,
+      ) || { name: defaultName, extension: defaultExtension, content: "" };
+
+      const otherComponents = project.components?.filter(
+        (component) =>
+          component.name !== defaultName &&
+          component.extension !== defaultExtension,
+      );
+
+      const updatedProject = {
+        ...project,
+        components: [
+          ...otherComponents,
+          {
+            ...mainComponent,
+            content: code,
+          },
+        ],
+      };
+      dispatch(setProject(updatedProject));
+    }
+  }, [code, loading]);
 };
--- a/src/hooks/useProject.js
+++ b/src/hooks/useProject.js
@@ -102,6 +102,7 @@ export const useProject = ({
 
       const updatedProject = {
         ...project,
+        project_type: project.project_type || "python",
         components: [
           ...otherComponents,
           {
--- a/src/hooks/useProject.js
+++ b/src/hooks/useProject.js
@@ -6,6 +6,7 @@ import { defaultPythonProject } from "../utils/defaultProjects";
 import { useTranslation } from "react-i18next";
 
 export const useProject = ({
+  assetsIdentifier = null,
   projectIdentifier = null,
   code = null,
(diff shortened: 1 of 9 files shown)


src/utils/apiCallHandler.js
```diff
@@ -69,6 +69,14 @@ export const readProject = async (projectIdentifier, locale, accessToken) => {
   );
 };
 
+export const loadAssets = async (assetsIdentifier, locale, accessToken) => {
+  const queryString = locale ? `?locale=${locale}` : "";
+  return await get(
+    `${host}/api/projects/${assetsIdentifier}/images${queryString}`,
```
conorriches, line 75: I'm unfamiliar with the admin side, will the query string ever change the assets returned? e.g. can we have a Spanish locale image vs an English locale image?

loiswells97: I guess potentially, maybe if there's text in there or something, or they want to provide a more culturally relevant image? The locale query string makes it load up a different project anyway, so it is possible to do this.

src/hooks/useProject.js
```diff
@@ -66,4 +81,36 @@ export const useProject = ({
     const data = defaultPythonProject;
     dispatch(setProject(data));
   }, [projectIdentifier, cachedProject, i18n.language, accessToken]);
+
+  useEffect(() => {
+    if (code && loading === "success") {
+      const defaultName = project.project_type === "html" ? "index" : "main";
+      const defaultExtension = project.project_type === "html" ? "html" : "py";
```
conorriches, line 88: I like this dynamic switching, but I think I'm not understanding something - if we only load assets, there will presumably be no `project` and thus no `project_type` which should therefore default to python.

Is there a way to set a default project type e.g. via props?

conorriches (commented):
Just a thought or two, but looking good so far

src/hooks/useProject.js
loiswells97: Yes, I'm sure we could do that, although I left it out for now as I'm not sure it's needed for this. I put the switching in so it can override a project set by the identifier as well (so where main/index already exists)

src/redux/EditorSlice.js
```diff
@@ -11,19 +11,24 @@ import {
...
-      { project, identifier, locale, accessToken, autosave },
+      { project, identifier, locale, accessToken, autosave, assetsOnly },
       { rejectWithValue },
     ) => {
       let response;
       switch (actionName) {
         case "load":
-          response = await readProject(identifier, locale, accessToken);
+          if (assetsOnly) {
+            response = await loadAssets(identifier, locale, accessToken, true);
```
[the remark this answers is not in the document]
loiswells97: Yes, you're right, not sure where that rogue `true` came from 😅

src/hooks/useProject.js
conorriches: Roger!

conorriches (approved):

conorriches (approved):

PR #915, opened 2024-01-30
Pass back more error information to output only component, hide e…

danhalson:
Ready for review

### DIFF
--- a/src/components/WebComponentProject/WebComponentProject.jsx
+++ b/src/components/WebComponentProject/WebComponentProject.jsx
@@ -7,7 +7,12 @@ import MobileProject from "../Mobile/MobileProject/MobileProject";
 import Output from "../Editor/Output/Output";
 import { defaultMZCriteria } from "../../utils/DefaultMZCriteria";
 import Sk from "skulpt";
-import { setIsSplitView, setWebComponent } from "../../redux/EditorSlice";
+import {
+  setIsSplitView,
+  setWebComponent,
+  setIsOutputOnly,
+  setErrorDetails,
+} from "../../redux/EditorSlice";
 import { MOBILE_MEDIA_QUERY } from "../../utils/mediaQueryBreakpoints";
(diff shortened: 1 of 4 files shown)


src/components/WebComponentProject/WebComponentProject.jsx
```diff
@@ -38,8 +44,7 @@ const WebComponentProject = ({
 
   dispatch(setIsSplitView(false));
   dispatch(setWebComponent(true));
-
-  console.log(`outputOnly: ${outputOnly}`);
+  dispatch(setIsOutputOnly(outputOnly));
```
conorriches, line 47: Apologies if my brain is scrambled 🍳  - it looks to me that calling `setIsOutputOnly` will always cause the [code here (line 197) ](https://github.com/RaspberryPiFoundation/editor-ui/pull/915/files#diff-3db14236c0b7d79bb1a84a935f0d541508cd0693a1c6e068faf3146420089504R197) to always set it to true even if the argument is false?

src/components/WebComponentProject/WebComponentProject.jsx
```diff
@@ -57,14 +62,14 @@ const WebComponentProject = ({
...
-      );
+
+      const payload = outputOnly
+        ? { errorDetails }
+        : { isErrorFree: error === "", ...mz_criteria };
+
+      document.dispatchEvent(runCompletedEvent(payload));
     }
-  }, [codeRunTriggered, codeHasRun, error]);
+  }, [codeRunTriggered, codeHasRun]);
```
conorriches, line 72: Hmm, see the GithubActions linter has raised a concern about the missing dependencies here

conorriches (commented):
Looking splendiferous! Brain might be scrambled but I have two teensy thoughts, otherwise think this is great 🥳

conorriches (approved):

PR #922, opened 2024-02-19
Fixing text output width on firefox

loiswells97:
## Before

Screenshot 2024-02-19 at 09 50 42

## After

Screenshot 2024-02-19 at 09 51 07


sra405 (approved):
👍

PR #964, opened 2024-03-12
Tweaks to Web Component

magdalenajadach:
Closing: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/138


src/containers/WebComponentLoader.jsx
```diff
@@ -17,6 +17,7 @@ import ToastCloseButton from "../utils/ToastCloseButton";
 
 import internalStyles from "../assets/stylesheets/InternalStyles.scss";
 import externalStyles from "../assets/stylesheets/ExternalStyles.scss";
+import webComponentStyles from "../assets/stylesheets/WebComponent.scss";
```
loiswells97, line 20: Can this not be added to the internal styles?

loiswells97:
Looks like some of the styles are still missing, for example, the coloured underline beneath selected tab, the colours for the selected tab in the sidebar, download icon in dark mode, settings ticks in dark mode against the teal background should be black. Also, if we're having the default theme colour being teal, then I think we need to think about whether the buttons need to default to teal as well rather than blue 🤔 
Screenshot 2024-03-15 at 11 29 32

src/containers/WebComponentLoader.jsx
magdalenajadach: Merged it one file, thank you.

magdalenajadach:
@loiswells97 Those findings are all known. I should flag it before the review. So the issue is with theming between Projects Site embedded WC, standalone WC and Editor. When applying this teal color we encountered loads of issues with styling and themes. Since we don't yet have a very clear vision of dark mode for the standalone WC we decided to push it as it is, since the Standalone in a light (default) mode was not usable for online courses. 
So long story short it's an update and progress, but not yet for the dark mode due to theming shenanigans.

loiswells97:
> @loiswells97 Those findings are all known. I should flag it before the review. So the issue is with theming between Projects Site embedded WC, standalone WC and Editor. When applying this teal color we encountered loads of issues with styling and themes. Since we don't yet have a very clear vision of dark mode for the standalone WC we decided to push it as it is, since the Standalone in a light (default) mode was not usable for online courses. So long story short it's an update and progress, but not yet for the dark mode due to theming shenanigans.

Ah I see, in that case, I'm happy to approve - as you say, it's an improvement that should make it usable in light mode for the online courses at least 👍

loiswells97 (dismissed):
Let's get this out there 🚢

magdalenajadach:
I managed to fix most of the issues with dark mode, removed peasky teal.

I achieved that overrriding some styles within InternalStyles.scss - it's not impacting Editor or Projectst site.

Still left some commenst to flag up what I have done.

magdalenajadach:
Screenshot 2024-03-18 at 13 27 20
Screenshot 2024-03-18 at 13 27 38

PR #985, opened 2024-04-17
Enable allowed external iframe src's

sra405:
As part of project migration from trinket to the code editor some additional functionality is required. This PR enables iframes to be used from the same approved external sources list as used for href's and other src attributes. By expanding this to allow for self references (i.e. src with a value the same as the `PUBLIC_URL`) users can embed other projects they (or we) own in iframes into their projects


sra405:
Allows iframes to be embedded if they're on the same domain as the editor or on `rpf.io`
image

loiswells97 (approved):
Looks great 👍

PR #990, opened 2024-04-25
Dispatch event when project identifier changes

floehopper:
See [this issue][1].

The integrated version of the editor in `editor-ui` redirects to the new project identifier URL after a project is remixed. However, the `editor-standalone` version which uses the editor web component was unable to detect changes in the project identifier and so the redirect was not working.

I've addressed this by making a new custom event ("editor-projectIdentifierChanged") available from the web component with the idea that we'll add an event listener in `ProjectComponentLoader` which redirects to the remixed project identifier URL (see https://github.com/RaspberryPiFoundation/editor-standalone/pull/63). The custom event is not dispatched when the project identifier is not set, because that's not currently useful in `editor-standalone`, but it would be easy enough to change.

This change should not change the behaviour of the `projects-ui` app which also uses the web component, because it is not listening for the new custom event. However, it would be easy enough to change that if it was desirable.

Note that I've added a separate integration test for `WebComponentProject`, because I wanted to use a real Redux store rather than the mock store that's being used in the unit test. This is the approach recommended [in the Redux docs][2]:

> Our recommendation for testing Redux-connected React components is via
> integration tests that include everything working together, with
> assertions aimed at verifying that the app behaves the way you expect
> when the user interacts with it in a given manner.

I hope the directory/file-naming convention I've used makes sense: `src/components/WebComponentProject/WebComponentProject.integration.test.js`.

[1]: https://github.com/RaspberryPiFoundation/editor-standalone/issues/55
[2]: https://redux.js.org/usage/writing-tests#integration-testing-connected-components-and-redux-logic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Support to enable embedding iframes in HTML projects from in-house domains (#985)
 - Unit tests for `pyodide` runner (#976)
+- Dispatch event when project identifier changes, e.g. after project is remixed (#2830)
 
 ## [0.22.2] - 2024-03-18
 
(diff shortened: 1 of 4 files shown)


sra405 (approved):
🙌 great stuff! Nice addition of integration tests too 🥳 happy with that convention for sure

PR #992, opened 2024-04-26
Add `load-remix-enabled` attribute to web component

floehopper:
This work came out of [this issue][1].

These changes mean that it is now possible to configure the web component not to load a logged-in user's remixed version of the project by setting `load_remix_disabled` to `true`.

We want to use this in `editor-standalone` with the web component editor so that we reproduce the behaviour currently in `editor-ui` with the integrated editor.

When the new attribute is set to `true`, the `loadRemix` property passed to `useProject` in `WebComponentLoader` is forced to be `false` so that the remixed version of the project will not be loaded.

The name of the new attribute isn't ideal - I would rather have used `load_remix_enabled` - however, boolean attributes of the web component default to `false` and I want the default behaviour to remain unchanged to avoid breaking the `projects-ui` app.

I've opened [this draft PR](https://github.com/RaspberryPiFoundation/editor-standalone/pull/85) on `editor-standalone` to illustrate the changes I plan to make. I'm hoping to add some end-to-end/integration tests in that repo to verify the new behaviour.

Screencast showing the fix in action in my local version of `editor-standalone` with the new `load_remix_disabled` attribute of the editor web component set to `true` in the `Project` & `MobileProject` components.

https://github.com/RaspberryPiFoundation/editor-ui/assets/3169/3f94dc06-f828-49f0-95f4-40848ec234bc

[1]: https://github.com/RaspberryPiFoundation/editor-standalone/issues/55.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Dispatch event when project identifier changes, e.g. after project is remixed (#2830)
 - Remove broken `format` script (#991)
 - Invalidate cached project when project is remixed (#1003)
+- Add `load_remix_disabled` attribute to web component (#992)
 
 ## [0.22.2] - 2024-03-18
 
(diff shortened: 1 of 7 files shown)


conorriches (approved):
Nice work!

chrislo (approved):

PR #1004, opened 2024-05-09
Release v0.23.0

floehopper:
I've grouped the unreleased changelog items into "Added", "Changed" & "Fixed" - I'm fairly confident that what I've done makes sense, but it would be worth another pair of eyes!

I've made this a minor version bump, because it includes both new functionality added in a backwardly compatible manner and backward compatible bug fixes, but no incompatible API changes as far as I'm aware. Please let me know if the latter is incorrect!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+## [0.23.0] - 2024-05-09
+
 ### Added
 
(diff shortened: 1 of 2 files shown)


sra405 (approved):
👍 lgtm

floehopper:
FYI I missed [the instructions in the README](https://github.com/RaspberryPiFoundation/editor-ui/blob/main/README.md#release-process) about how to do this - so I've fixed a couple of things in https://github.com/RaspberryPiFoundation/editor-ui/pull/1005.

Issue #986, opened 2024-04-17
Edit project name in web component


sra405:
Given we want to use the web component in editor-standalone and switch editor.raspberrypi.org traffic over to that repo we need to ensure the existing functionality is supported in the web component. One aspect that needs including is the ability to edit a projects name in the web component.

[Slack thread](https://raspberrypifoundation.slack.com/archives/C02JBAA2NFP/p1713352569352239)

### Notes

- Some web component integrations need the existing read-only project name functionality (i.e the [projects site integration](https://projects.raspberrypi.org/en/projects/editor-hello-world/editor))
- Consider toggling this functionality using an attribute to `editor-wc`

floehopper:
It's worth noting that it is already possible to rename a project on the "My projects" page; just not in the project bar of the editor web component.

floehopper:
Hopefully addressed in https://github.com/RaspberryPiFoundation/editor-ui/pull/1009 & https://github.com/RaspberryPiFoundation/editor-standalone/pull/125.

MFarringtonRPF:
@danhalson can this be marked as closed given comments above? Looks to be added as well in the temp branch.

MFarringtonRPF:
Unclear if still needed.

floehopper:
@MFarringtonRPF I'm pretty sure it's OK to close this - the changes I made in those two PRs mentioned above allow users to rename a project by clicking on the project name in editor-standalone.

PR #1009, opened 2024-05-15
Add project_name_editable attribute to web component

floehopper:
Addresses #986 which is part of RaspberryPiFoundation/editor-standalone#16.

This will allow us to make the project name editable in `editor-standalone` (see RaspberryPiFoundation/editor-standalone#125) while leaving other projects using the web component (e.g. `projects-ui`) and the integrated version in `editor-ui` unchanged.

Note that this new attribute only affects the `ProjectName` component in the `ProjectBar` and not the one in the sidebar, i.e. in `ProjectsPanel`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- Add `project_name_editable` attribute to web component (#1009)
+
 ### Changed
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ The web component can be included in a page by using the `<editor-wc>` HTML elem
 - `code`: A preset blob of code to show in the editor pane.
 - `sense_hat_always_enabled`: Show the Astro Pi Sense HAT emulator on page load
 - `load_remix_disabled`: Do not load a logged-in user's remixed version of the project specified by `identifier` even if one exists (defaults to `false`)
+- `project_name_editable`: Allow the user to edit the project name in the project bar (defaults to `false`)
 
 ### `yarn start:wc`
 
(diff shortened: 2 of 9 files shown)


loiswells97 (commented):
This looks great overall. My only comment is that the variable name for the boolean that determines whether the name is editable changes throughout (`projectNameEditable -> nameEditable -> editable`). Maybe this is intentional though?

floehopper:
> My only comment is that the variable name for the boolean that determines whether the name is editable changes throughout (`projectNameEditable -> nameEditable -> editable`). Maybe this is intentional though?

Yes, it was intentional - I felt as if the context is different in each case - see [Remove redundant prefixes from projectNameEditable](https://github.com/RaspberryPiFoundation/editor-ui/pull/1009/commits/19886e62e5af1ec3e2bb9abd81ef5267404237cb). It doesn't make much sense to me to use a property called `projectNameEditable` in the `ProjectName` component - `editable` makes more sense to me.

loiswells97:
> > My only comment is that the variable name for the boolean that determines whether the name is editable changes throughout (`projectNameEditable -> nameEditable -> editable`). Maybe this is intentional though?
> 
> Yes, it was intentional - I felt as if the context is different in each case - see [Remove redundant prefixes from projectNameEditable](https://github.com/RaspberryPiFoundation/editor-ui/pull/1009/commits/19886e62e5af1ec3e2bb9abd81ef5267404237cb), e.g. it doesn't make much sense to me to use a property called `projectNameEditable` in the `ProjectName` component - `editable` makes more sense to me, because the `projectName` prefix is redundant.

Okay, fair enough, I think that makes sense 👍 Let's leave it as it is in that case

loiswells97 (approved):
Looks great, let's get this out there! 👍

PR #1013, opened 2024-05-17
Remove unused `/embedded/projects/:identifier` route

floehopper:
As far as we can tell, this route is [no longer used anywhere][1].

This was originally added in Mar 2022 in #72 for the "ESA demo". I suspect we now achieve a similar effect by using the web component with [its `embedded` attribute](https://github.com/RaspberryPiFoundation/editor-ui/blob/340ab1192fb3358ad4614034208603948f8b95d6/src/web-component.js#L50) set to `true`, rather than using an `iframe` with its `src` set to this route.

I found it confusing to see both the `/:locale/embed/viewer/:identifier` and this route in the `AppRoutes` component. Removing this one makes the code less confusing.

I've also removed the now redundant `embedded` property on the `ProjectComponentLoader` component and the similarly redundant call to `useEmbeddedMode` which is a no-op when `embedded` is `false`.

[1]: https://github.com/search?q=org%3Araspberrypilearning+embedded%2Fprojects&type=code

### DIFF
--- a/src/containers/ProjectComponentLoader.jsx
+++ b/src/containers/ProjectComponentLoader.jsx
@@ -1,7 +1,6 @@
 import React, { useEffect } from "react";
 import { useSelector } from "react-redux";
 import { useProject } from "../hooks/useProject";
-import { useEmbeddedMode } from "../hooks/useEmbeddedMode";
 import { useMediaQuery } from "react-responsive";
 import { useNavigate, useParams } from "react-router-dom";
 import { useTranslation } from "react-i18next";
@@ -21,7 +20,6 @@ import { useProjectPersistence } from "../hooks/useProjectPersistence";
 const ProjectComponentLoader = (props) => {
   const loading = useSelector((state) => state.editor.loading);
   const { identifier } = useParams();
-  const embedded = props.embedded || false;
   const user = useSelector((state) => state.auth.user);
   const accessToken = user ? user.access_token : null;
   const project = useSelector((state) => state.editor.project);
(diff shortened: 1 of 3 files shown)


src/containers/ProjectComponentLoader.jsx
```diff
@@ -21,7 +20,6 @@ import { useProjectPersistence } from "../hooks/useProjectPersistence";
 const ProjectComponentLoader = (props) => {
   const loading = useSelector((state) => state.editor.loading);
   const { identifier } = useParams();
-  const embedded = props.embedded || false;
```
loiswells97, line 24: Might be missing something, but why is this now redundant?

sra405 (dismissed):
👍 nice, definitely simplifies the loader!

src/containers/ProjectComponentLoader.jsx
floehopper: > Might be missing something, but why is this now redundant?

The route that I removed seemed to be the only place where the `embedded` property for `ProjectComponentLoader` was set to `true`. The `embedded` property itself defaults to `false` and thus after the route was removed the value of the property must always be `false`. The only thing the `embedded` proeprty was used for was to call `useEmbeddedMode()`. However, that function is a no-op when `embedded` is `false`. So we can remove all of it. Does that make sense?

floehopper: In case it helps, there is still an `embedded` attribute on the web component: https://github.com/RaspberryPiFoundation/editor-ui/blob/340ab1192fb3358ad4614034208603948f8b95d6/src/web-component.js#L50 which is still used in `WebComponentLoader`.

loiswells97: Ah I see, yes, that makes sense - I was worried about some of the embedded viewer stuff, but you're right, that's not using the `ProjectComponentLoader` so it should be fine 👍

loiswells97 (approved):
Looks good 👍

PR #1016, opened 2024-05-21
Remove unused `isEmbedded` param from `useProject` call in `EmbeddedViewer`

floehopper:
In [this commit][1], the `isEmbedded` param for the `useProject` hook was removed, but `EmbeddedViewer` is still passing it in. The `isEmbedded` value is obtained from the Redux store instead.

[1]: https://github.com/RaspberryPiFoundation/editor-ui/commit/28ee44fbe671d84fd646be12b0e28c25c1da6194


chrisroos (approved):
LGTM :+1:

Issue #1010, opened 2024-05-17
Productionise Output only Web Component


sra405:
Similar to the embedded view from the site I wanted to see if it was simple enough to add an output only view to the web component.

It's worth noting the need for this in two use cases:
- [Productionising Block-to-text](https://github.com/RaspberryPiFoundation/projects-ui/issues/2498) - we need the ability to pass in code via the `code` attribute for users to run per step in the activity - draft PR used as the code output runner in the alpha [here](https://block-to-text-alpha.pages.dev/exercise/example-project). As a draft this was sufficient for the alpha, for use in a beta (integrated with the projects site) we'll need to add tests to ensure the functionality works as expected
- [Editor web component separation](https://github.com/RaspberryPiFoundation/editor-standalone/issues/16) - as part of splitting out the editor.raspberrypi.org app and the web component the need to have an output only web component enables the embedded path in the app for integration via iframes
  - the editor currently has a `embed/viewer` path (e.g. [here](https://editor.raspberrypi.org/en/embed/viewer/welcome-to-antarctica-step2)) for use in iframes in external products (e.g. [here](https://projects.raspberrypi.org/en/projects/welcome-to-antarctica/1)) - we'll need this to properly separate out the editor-standalone by utilising an output only web component on this path
  - all functionality in editor-ui should be in the [EmbeddedViewer component](https://github.com/RaspberryPiFoundation/editor-ui/blob/main/src/components/EmbeddedViewer/EmbeddedViewer.jsx)
  - additional functionality over just running code is needed, this includes:
    - ability to obtain code/assets from a project identifier
    - inclusion of a file bar at the top of the page (with the ability to choose an alternative file to run)
    - inclusion of a built in run CTA rather than an external trigger

### Notes

If its easier to do this in parts I'm all ears. The functionality for block to text is there and mostly aligns with what's needed for the standalone. Additionally if we think we can provide the functionality of the embedded viewer path as a web component I'm all ears too, we'd significantly reduce our number of iframe (not to zero unfortunately) but would also unlock performance gains too!

floehopper:
@sra405 I've created [this draft PR](https://github.com/RaspberryPiFoundation/editor-ui/pull/1019) by doing the following:

* Curating the commits from #782 on top of the latest code in the `main` branch
* Checking this new branch works in conjunction with [the Block-to-Text app](https://github.com/RaspberryPiFoundation/block-to-text-alpha/) running locally
[...]

PR #1019, opened 2024-05-22
Add support for Block-to-Text app

floehopper:
### Summary

This adds a bunch of new attributes (e.g. `output_only`, `output_panels` & `assets_identifier`), methods (e.g. `rerunCode`) and behaviour to the web component to support [the Block-to-Text app](https://block-to-text-alpha.pages.dev/).

This branch is heavily based on the work in #782 which is what is what is currently being used by the alpha version of the Block-to-Text app. I've cherry-picked commits from the original work onto the latest work in`main` and curated the commits a bit to try to make them atomic and easier to follow.

Screenshot 2024-05-31 at 10 25 39

### Detail

In each commit:

* I've reviewed the code
* I've added unit tests
* I've fixed any bugs that I've found
* I've tried to explain the change at a high level
* I've tried to explain the motivations behind the change

Obviously writing the explanations after the fact and not being the original author of the changes is sub-optimal, so please let me know if I've got anything wrong!

In the final commit, I've also added a new Cypress end-to-end spec which uses the existing `src/web-component.html` page to configure the web component in a similar way to how the Block-to-Text app uses it. The spec uses the web component to runs a simple P5 sketch, drawing some text and an image from an existing project. Hopefully this gives us some confidence that the web component will work OK with the Block-to-Text app.

332883390-ddd6800a-2ce1-4278-b7e3-8129604c1153

### Notes/Questions

I haven't included [the change](https://github.com/RaspberryPiFoundation/editor-ui/pull/782/files#diff-d355fe835fa320ad389b053fa105e008bf035de5f236519bb92264fb061c04c9R65) which exposes the `embedded` property as a web component attribute, even though the Block-to-Text app is [setting it](https://github.com/RaspberryPiFoundation/block-to-text-alpha/blob/d16a8ecc1d3da202363fb16db31f4bf93f46aa04/src/components/Editor.js#L21), because I don't believe it's necessary. Please let me know if you think I'm mistaken!

I do have some concerns about the number of configuration options knocking around in the code, particularly all the boolean ones where I suspect there may well be some overlap. I understand why it's like this, but I don't think it helps us that some configuration is done via properties, some by query string params, and some by web component attributes. Also I'm not entirely convinced that configuration options belong in a Redux store. However, I am hopeful that it will be easier to rationalize all this a bit in [the 2nd step](https://github.com/RaspberryPiFoundation/editor-standalone/issues/16) of [this issue](https://github.com/RaspberryPiFoundation/editor-ui/issues/1010) and it seems a bit too risky to do any of that at this stage.

I also wonder whether we could simplify/reduce the number of web component attributes, e.g. perhaps we could have `identifier` & `assetsOnly` instead of `identifier` & `assetsIdentifier`; or perhaps we could even have a higher-level `mode` attribute which had a special "Block-to-Text" value which set the relevant lower-level properties...? However, I think we should probably just get the changes in this PR merged and tackle any of that later.

I've run the Block-to-Text app locally against this branch of `editor-ui` and worked my way through all the steps of the exercise successfully. However, I haven't done a great deal of manual testing of the `editor-ui` app or of the web component in other modes - I'm hoping that the existing automated tests will have highlighted any regressions. Let me know if you think I should do some manual testing of particular functionality before merging.

Closes https://github.com/RaspberryPiFoundation/projects-ui/issues/2498.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add `output_only` attribute to web component (#1019 & originally #782)
 - Add `assets_identifier` attribute to web component (#1019 & originally #901)
 - Enhance `code` attribute on web component to override project main component content (#1019 & originally #901)
+- Add `runCode`, `stopCode` & `rerunCode` methods to web component (#1019 & originally #899)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fires custom event when the theme changes (#1015)
 - Add `output_only` attribute to web component (#1019 & originally #782)
 - Add `assets_identifier` attribute to web component (#1019 & originally #901)
+- Enhance `code` attribute on web component to override project main component content (#1019 & originally #901)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Enhance `code` attribute on web component to override project main component content (#1019 & originally #901)
 - Add `runCode`, `stopCode` & `rerunCode` methods to web component (#1019 & originally #899)
 - Send error details in "editor-runCompleted" event (#1019 & originally #915)
+- Return error details to web component (#1019 & originally #915)
+- Add `output_panels` attribute to web component (#1019 & originally #909)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add `project_name_editable` attribute to web component (#1009)
 - Fires custom event when the theme changes (#1015)
 - Add `output_only` attribute to web component (#1019 & originally #782)
+- Add `assets_identifier` attribute to web component (#1019 & originally #901)
 
 ### Changed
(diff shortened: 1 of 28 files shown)


chrislo (commented):
Minor: In the commit note to 12b3ef6a I found the phrase `Previously it was typically not include any other components` a bit hard to understand.

chrislo:
This all looks good to me, in particular the addition of the e2e test in the final commit really helped me understand how everything hangs together. 

> I also wonder whether we could simplify/reduce the number of web component attributes, e.g. perhaps we could have identifier & assetsOnly instead of identifier & assetsIdentifier; or perhaps we could even have a higher-level mode attribute which had a special "Block-to-Text" value which set the relevant lower-level properties...? However, I think we should probably just get the changes in this PR merged and tackle any of that later.

I agree with this. It feels like if there were a way to reduce the number of arguments (via a combined `mode` flag, or perhaps a higher level component that sets the properties required for block to text) it might make it easier to understand what's happening at the call site (and discourage future us from adding additional options). Having said that I don't think that should stop you merging things now. 

> Obviously writing the explanations after the fact and not being the original author of the changes is sub-optimal, so please let me know if I've got anything wrong!

I wasn't really able to help with that - I think it might be worth @sra405, @loiswells97 and/or @danhalson reading through the individual commit messages (rather than the combined diff) at least to see if there's any useful context missing that we can add to the commit notes to help our future selves. But I'm going to approve this for now so that you're not blocked 👍

chrislo (approved):

floehopper:
> Minor: In the commit note to [12b3ef6](https://github.com/RaspberryPiFoundation/editor-ui/commit/12b3ef6af6513ce3f000efcdaf8f394911b96ec4) I found the phrase `Previously it was typically not include any other components` a bit hard to understand.

@chrislo I'm sorry - I somehow missed this comment before I merged - hopefully it's not _too_ confusing!

PR #1020, opened 2024-05-23
Add e2e spec for project remix behaviour in web component

floehopper:
The `editor-standalone` app will:

* Listen for "editor-projectIdentifierChanged" custom events from the web component and will redirect to the remixed project's URL. The relevant web component behaviour was changed in #990. The relevant `editor-standalone` behaviour was changed in RaspberryPiFoundation/editor-standalone#63.
* Set the web component `load_remix_disabled` attribute to `true` so that the original project will be loaded in preference to the remixed version (unlike the way projects-ui works). The relevant web component behaviour was changed in #992 & #1003. The relevant `editor-standalone` behaviour will be changed in RaspberryPiFoundation/editor-standalone#85.

This spec checks that the web component provides the relevant behaviour for editor-standalone to use.

I did attempt to add an example for the default behaviour (i.e. when `load_remix_disabled` is not set). However, without any way to easily reset the state of the projects in the editor-api database, it's hard to come up with a robust way to write such a test.

Note that this PR changes both the editor app itself and the fake "app" wrapping the web component to use https://test-editor-api.raspberrypi.org instead of https://staging-editor-api.raspberrypi.org in CI. The former has `BYPASS_AUTH` set to `true` so the apps can make requests as if the user was logged in. The new e2e spec also relies on the `projects:create_all` rake task having been run - which we've only done manually for now, but needs automating.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -94,7 +94,7 @@ jobs:
           wait-on: "http://localhost:3000, http://localhost:3001"
           quiet: true
         env:
-          REACT_APP_API_ENDPOINT: "https://staging-editor-api.raspberrypi.org"
+          REACT_APP_API_ENDPOINT: "https://test-editor-api.raspberrypi.org"
           PUBLIC_URL: "http://localhost:3000"
           ASSETS_URL: "http://localhost:3000"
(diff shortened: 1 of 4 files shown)


chrisroos (approved):
Nice work! :+1:

PR #1021, opened 2024-05-28
Fix initial value of `user` in `WebComponentLoader`

floehopper:
Previously, even if the `authKey` property was provided and a logged-in user was stored in `localStorage`, on first render the value of `user` was `null`. This was because the `localStorageUserMiddleware` which dispatches `setUser` to store the user in Redux had not yet been triggered.

**This meant that previously if we tried to use the web component to view a project that the logged-in user owned, the request to editor-api to load the project would respond with a 403 Forbidden and the project would not load.**

I'm fairly (but not completely) sure that the assertions in the "with user set in local storage" group of examples in
`src/containers/WebComponentLoader.test.js` were incorrect, i.e. if the `authKey` is set and a `user` is present in `localStorage` we ought to load the project based on the fact the user is logged in. I have fixed these assertions.

The `null` value of the `user` property in the call to `useProjectPersistence` in the "When no user is in state with no user in local storage" group of examples in `src/containers/WebComponentLoader.test.js` is due to the return value for `localStorage.getItem(authKey)` and not the fallback in the new ternary expression in `WebComponentLoader`. I think this is fine.

I'm not sure whether my fix is definitely correct/idiomatic - maybe I should be using a `useState` hook for the `localStorageUser` in combination with a `useEffect` hook...?

Also stepping back a bit, it's not entirely clear to me that [dispatching `setUser`][1] in `localStorageUserMiddleware` is definitely the most sensible place to do that.

[1]: https://github.com/RaspberryPiFoundation/editor-ui/blob/92b45d59d8187b26538c03af825bb34313b7e344/src/redux/middlewares/localStorageUserMiddleware.js#L11

### DIFF
--- a/src/containers/WebComponentLoader.jsx
+++ b/src/containers/WebComponentLoader.jsx
@@ -46,7 +46,10 @@ const WebComponentLoader = (props) => {
   const { t } = useTranslation();
   const [projectIdentifier, setProjectIdentifier] = useState(identifier);
   localStorage.setItem("authKey", authKey);
-  const user = useSelector((state) => state.auth.user);
+  const localStorageUser = authKey
+    ? JSON.parse(localStorage.getItem(authKey))
+    : null;
+  const user = useSelector((state) => state.auth.user || localStorageUser);
   const [loadCache, setLoadCache] = useState(!!!user);
   const [loadRemix, setLoadRemix] = useState(!!user);
   const project = useSelector((state) => state.editor.project);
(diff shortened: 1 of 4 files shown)


src/containers/WebComponentLoader.jsx
```diff
@@ -46,7 +46,10 @@ const WebComponentLoader = (props) => {
   const { t } = useTranslation();
   const [projectIdentifier, setProjectIdentifier] = useState(identifier);
   localStorage.setItem("authKey", authKey);
-  const user = useSelector((state) => state.auth.user);
+  const localStorageUser = authKey
```
danhalson, line 49: (regarding your comment about this) I think this is fine since authKey is passed in as an argument to the component, and can only be updated on render

danhalson (approved):
looks great..I do agree about dispatching setUser in middleware, but I'm not sure we need to address that now?

src/containers/WebComponentLoader.jsx
floehopper: Great! 😎

PR #1022, opened 2024-05-28
Make `authKey` in e2e web component spec more realistic

floehopper:
The comment about the `authKey` needing to match `HydraPublicApiClient::BYPASS_AUTH_USER_ID` in editor-api was incorrect. While the actual value is unimportant in this spec, it seems better to make it as realistic as possible.


chrisroos (dismissed):
LGTM :+1:

PR #1023, opened 2024-05-28
Remove unused `ComponentStore`

floehopper:
This was introduced in #96, but never used. I suspect the `import store from "./app/store"` statement in `src/web-component.js` was intended to import it, but it used the wrong filename.

`ComponentStore` was later effectively replaced by `WebComponentStore` in #728.


chrisroos (approved):
LGTM :+1:

PR #1030, opened 2024-06-04
Support use of the web component in editor-standalone's embedded view

floehopper:
This makes a number of changes to support the use of the web component in the embedded view of `editor-standalone`. See [this PR](https://github.com/RaspberryPiFoundation/editor-standalone/pull/164) for details.

In particular, I've tried to "tighten" up what we mean by the `isEmbedded` state (and its new corresponding `embedded` web component attribute). It's now very specifically intended to indicate that the we are in the embedded view i.e. on [the `/:locale/embed/viewer/:identifier` route](https://github.com/RaspberryPiFoundation/editor-standalone/blob/b38a307ea78ae672593befa0875fb550c9ba6689/src/components/AppRoutes.jsx#L75-L78).

In conjunction with [this block-to-text PR](https://github.com/RaspberryPiFoundation/block-to-text-alpha/pull/125), this has allowed me to remove some checks on the `webComponent` state which didn't completely make sense. This `webComponent` state now only indicates that we are in the context of the web component and should *not* be used as a proxy for e.g. whether we are in the context of the Block-to-Text app. There are some other usages of the `webComponent` state within the web component code, but I'm fairly confident they all still make sense.

As well as allowing us to use the web component for the embedded view in `editor-standalone`, the new `output_split_view` web component attribute will allow us to fix a minor difference in behaviour of the editor itself in `editor-standalone` versus `editor-ui`.

The changes are as follows:

* [Add boolean embedded attribute to web component](https://github.com/RaspberryPiFoundation/editor-ui/pull/1030/commits/7092ec6d10b367f05128e78053ffdd7bbe43e1a1)
* [Remove embedded & browserPreview props from Output](https://github.com/RaspberryPiFoundation/editor-ui/pull/1030/commits/c18d8f888786527c6a4ecd69ae6dce7ef9a381c5)
* [Ignore webComponent state for RunBar in Output](https://github.com/RaspberryPiFoundation/editor-ui/pull/1030/commits/73586cbe651ebd487158ec1498d89615f062dffb)
* [Ignore webComponent state for preview link in HtmlRunner](https://github.com/RaspberryPiFoundation/editor-ui/pull/1030/commits/a48eda50ac7f4fb1f551a18364d235532fea719c)
* [Add output_split_view attribute to web component](https://github.com/RaspberryPiFoundation/editor-ui/pull/1030/commits/1c4cdd241d7d01592e1123868c5252360bb09ff6)

See individual commit notes for details.

### Testing

I have tested these changes and those in RaspberryPiFoundation/editor-standalone#164 against the `editor-ui` app and the `editor-standalone` app and compared the behaviour against that currently deployed to [production](https://editor.raspberrypi.org):

#### Python project

* Editor: `/en/projects/<python-project-identifier>`
  * "Run" button in "editor-input" (below code)
  * No "Run" button in output panel
  * "Visual" & "Text" output panels (switchable between tabbed/split)
  * No preview link
* Embedded: `/en/embed/viewer/<python-project-identifier>`
  * Only output visible; no code/project
  * "Run" button at bottom of output
  * "Visual" & "Text" output panels (split view)

#### HTML project

* Editor: `/en/projects/<html-project-identifier>`
  * "Run" button in "editor-input" (below code)
  * No "Run" button in output
  * Preview of rendered HTML in output
  * Preview link opens in another browser tab
* Embedded: `/en/embed/viewer/<html-project-identifier>`
  * Only output visible; no code/project
  * "Run" button at bottom of output
  * Full page preview of rendered HTML in output (runs automatically)
* Preview: `/en/embed/viewer/<html-project-identifier>?browserPreview=true&page=index.html`
  * Only output visible; no code/project
  * No "Run" button at bottom of output
  * Full page preview of rendered HTML in output (runs automatically)
* Preview (other page): `/en/embed/viewer/<html-project-identifier>?browserPreview=true&page=other.html`
  * Only output visible; no code/project
  * No "Run" button at bottom of output
  * Full page preview of *other* rendered HTML in output (runs automatically)

Relates to #1010 & RaspberryPiFoundation/editor-standalone#16.

### DIFF
--- a/cypress/e2e/spec-wc.cy.js
+++ b/cypress/e2e/spec-wc.cy.js
@@ -142,3 +142,62 @@ describe("when load_remix_disabled is true, e.g. in editor-standalone", () => {
     });
   });
 });
+
+describe("when embedded, output_only & output_split_view are true", () => {
+  const urlFor = (identifier) => {
+    const params = new URLSearchParams();
+    params.set("identifier", identifier);
+    params.set("load_remix_disabled", "true");
+    params.set("embedded", "true");
+    params.set("output_only", "true");
+    params.set("output_split_view", "true");
+    return `${origin}?${params.toString()}`;
+  };
+
+  it("displays the embedded view for a Python project", () => {
+    cy.visit(urlFor("clean-car-example"));
+
+    // Check text output panel is visible and has a run button
+    // Important to wait for this before making the negative assertions that follow
+    cy.get("editor-wc").shadow().contains("Text output").should("be.visible");
+    cy.get("editor-wc").shadow().find("button").contains("Run").should("be.visible");
+
+    // Check that the side bar is not displayed
+    cy.get("editor-wc").shadow().contains("Project files").should("not.exist");
+    // Check that the project bar is not displayed
+    cy.get("editor-wc").shadow().contains("Don't Collide: Clean Car").should("not.exist");
+    // Check that the editor input containing the code is not displayed
+    cy.get("editor-wc").shadow().contains("# The draw_obstacle function goes here").should("not.exist");
+
+    // Run the code and check it executed without error
+    cy.get("editor-wc").shadow().find("button").contains("Run").click();
+    cy.get("#results").should("contain", '{"errorDetails":{}}');
+
+    // Check that the visual output panel is displayed in split view mode (vs tabbed view)
+    cy.get("editor-wc").shadow().contains("Visual output").should("be.visible");
+    cy.get("editor-wc").shadow().contains("Visual output").parents("ul").children().should("have.length", 1);
+  });
+
+  it("displays the embedded view for an HTML project", () => {
+    cy.visit(urlFor("anime-expressions-solution"));
+
+    // Check HTML preview output panel is visible and has a run button
+    // Important to wait for this before making the negative assertions that follow
+    cy.get("editor-wc").shadow().contains("index.html preview").should("be.visible");
+    cy.get("editor-wc").shadow().find("button").contains("Run").should("be.visible");
+
+    // Check that the code has automatically run i.e. the HTML has been rendered
+    cy.get("editor-wc").shadow().find("iframe#output-frame").its("0.contentDocument.body").should("contain", "Draw anime with me");
+
+    // Check that the side bar is not displayed
+    cy.get("editor-wc").shadow().contains("Project files").should("not.exist");
+    // Check that the project bar is not displayed
+    cy.get("editor-wc").shadow().contains("Anime expressions solution").should("not.exist");
+    // Check that the editor input containing the code is not displayed
+    cy.get("editor-wc").shadow().contains("<h1>Draw anime with me</h1>").should("not.exist");
+
+    // Run the code and check it executed without error
+    cy.get("editor-wc").shadow().find("button").contains("Run").click();
+    cy.get("#results").should("contain", '{"errorDetails":{}}');
+  });
+});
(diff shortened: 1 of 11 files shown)


src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -327,7 +326,7 @@ function HtmlRunner() {
                   "output.preview",
                 )}`}</span>
               </Tab>
-              {!!!isEmbedded && !!!webComponent && (
+              {!!!isEmbedded && (
```
conorriches, line 329: Slightly tangentially, I've not seen `!!!` before - surely this evaluates to just `!`. Is there a reason we're using it here?

conorriches (approved):

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
floehopper: @conorriches

> surely this evaluates to just `!`

You are correct, but only if the value of `isEmbedded` was guaranteed to always be `true` or `false`.

I think this is a variation of the `!!x` pattern. The latter converts `null` and `undefined` to `false`. Thus `!!!x` that's being used here converts `null` and `undefined` to `true`. Does that make sense?

As an aside, I'm not a big fan of this pattern, because I think it's often used *unecessarily* to be defensive against unexpected values and it makes the code hard to read. I think a better approach is to ensure that the value *never* gets set to `null` or `undefined` in the first place.

cypress/e2e/spec-wc.cy.js
```diff
@@ -142,3 +142,58 @@ describe("when load_remix_disabled is true, e.g. in editor-standalone", () => {
...
+    cy.get("editor-wc").shadow().contains("index.html preview").should("be.visible");
+    cy.get("editor-wc").shadow().find("button").contains("Run").should("be.visible");
+
+    // Check that the code has automatically run i.e. the HTML has been rendered
+    cy.get("editor-wc").shadow().find("iframe#output-frame").its("0.contentDocument.body").should("contain", "Draw anime with me");
+
+    // Check that none of the other editor components are displayed in output-only mode
+    cy.get("editor-wc").shadow().contains("Project files").should("not.exist");
+    cy.get("editor-wc").shadow().contains("Anime expressions solution").should("not.exist");
+    cy.get("editor-wc").shadow().contains("<h1>Draw anime with me</h1>").should("not.exist");
```
conorriches, line 193: I'm not so familiar with cypress but this stands out to me - only because it could be interpreted that it's searching for the string of `<h1>Draw anime with me</h1>` - do we want to be finding a `h1` element that contains the text `Draw anime with me`? Or does this effectively parse the string as a `h1` tag?

conorriches: Bonus thought - I'm mostly just concerned that this passes for the right reasons - if it's searching for the parsed string of `<h1>Draw anime...` it could possibly pass erroneously as we are checking for an absence which will pass much more readily.

floehopper: Ah, yes - I can see how that might be confusing. In this case I want to check the editor input component is present and displaying the code, so the text "<h1>Draw anime with me</h1>" is actually visible to the user including the "<h1>" bits! Maybe I should add a comment to explain that a bit...?

floehopper: @conorriches I've tried to address this in https://github.com/RaspberryPiFoundation/editor-ui/pull/1030/commits/58276a3cce45be82cdd5ffdc59bb48ebafd735e by adding comments like "Check that the editor input containing the code is not displayed". Does that help?

floehopper: > Bonus thought - I'm mostly just concerned that this passes for the right reasons - if it's searching for the parsed string of `<h1>Draw anime...` it could possibly pass erroneously as we are checking for an absence which will pass much more readily.

Yeah - I did check that it fails if I do not set `output_only` to `true`. Also I'd like to think there are other e2e specs that would fail if the source code was not displayed when `output_only` is not set.

conorriches: Fab - thanks @floehopper ! That clarifies it nicely

PR #1035, opened 2024-06-07
Fix web-component context

sra405:
Currently the web component doesn't have access to the browser context. This typically isn't an issue until we utilise components like `Link` which soft reloads pages and keeps state with a `useContext` hook.

https://github.com/RaspberryPiFoundation/editor-ui/assets/74183390/40ee68a3-287c-4c01-820b-1b90cc06bbcb

### Fix

I wasn't expecting wrapping the web component in a `BrowserRouter` to work but it does and means this issue is relatively trivial. I do suspect this comes with caveats and its worth being mindful there may be differences in the implementation in a web component we're not expecting


loiswells97 (approved):
Wahay, still can't quite believe this worked! 🤯 🎉

PR #1041, opened 2024-06-12
UAT web component fixes

danhalson:
Adds support to use built in editor styles in web component, currently defaults to general design system colours. If the flag is true, import those styles and applies them, else falls back to design system styling.


loiswells97 (commented):
Just had a look at the web component on the review app and the buttons look great! The lines under the selected tabs have disappeared though 🤔 The ticks in the settings sidebar tab are still blue, although I'm not sure how much we care about that one, but we should probably fix the selected tabs I guess

loiswells97 (approved):
Yes, this looks brilliant! 🤩 Let's get it out there!

PR #1042, opened 2024-06-13
Release 0-25-2

danhalson:
### Added

- Web component now supports using standard editor-ui styles if useEditorStyles is true

### Fixed

- Fix bug where `skulpt` runs the first component rather than `main.py` (#1039)
- Fixing bug with `HTML` projects loading after remix if they have multiple `HTML files` (#1040)


sra405 (approved):
👍

Issue #1044, opened 2024-06-19
Avoid using stale access tokens in requests to Editor API


chrisroos:
We've been seeing quite a few [`Faraday::Unauthorized` errors reported by Sentry in Editor API](https://rpf.sentry.io/issues/5135885959/). These errors are being caused by stale access tokens being sent in the requests from the WebComponent.

I've narrowed the problem down to [this line in WebComponentLoader](https://github.com/RaspberryPiFoundation/editor-ui/blob/e3495a24ff296cf0c3c22910db25d02547c6dfd9/src/containers/WebComponentLoader.jsx#L58) which was introduced in [this commit by James M](https://github.com/RaspberryPiFoundation/editor-ui/commit/373b7162ae26d038cbbd1213a3f68db1f289b531). James's fix works in the case where the user stored in localstorage has an active (i.e. not expired) access token but causes failures if the access token has expired. I contemplated fixing this by checking the expiry date of the access token in `WebComponentLoader` but that doesn't feel quite right because that must already be happening somewhere else in the app (maybe in a library we're using for oidc)? Instead, I think it might be better if the `WebComponentLoader` only reads the user from state (and not from localstorage), and we rely on the component re-rendering when the user becomes available in the state. We know that the `WebComponentLoader` component renders (and therefore makes a request to editor-api with no `Authorization` header) before the user is available in state. This is fine for public projects (i.e. those where `user_id` is null) but fails for user projects (i.e. those where `user_id` is set). When the user becomes available in state I'd expect the component to re-render and make a request to editor-api with the `Authorization` header set correctly. At present, this second request doesn't happen because of [the condition in this line of `syncProject`](https://github.com/RaspberryPiFoundation/editor-ui/blob/e3495a24ff296cf0c3c22910db25d02547c6dfd9/src/redux/EditorSlice.js#L65).

chrisroos:
I've been playing around with some ideas for this in PR #1043.

chrisroos:
I wonder whether it might be worth going with the quicker fix for now, of not loading the user from localstorage if the access token has expired, as I think that might help reduce the errors we're seeing in production?

floehopper:
As suggested by @chrisroos I've worked up the quick fix into https://github.com/RaspberryPiFoundation/editor-ui/pull/1046 to address this issue.

floehopper:
[...]

PR #1046, opened 2024-06-19
Quick fix for expired access token problem in web component

floehopper:
This is a quick fix for https://github.com/RaspberryPiFoundation/editor-ui/issues/1044.

We've been seeing quite a few [`Faraday::Unauthorized` exceptions reported by Sentry in Editor API][1]. These errors are being caused by expired access tokens being sent in the requests from the web component.

This can happen if a user is logged in and is viewing a non-public project, but they close their browser and return to the project some time (1 hour?) later when their access token has expired. In this scenario the project will not load, because of the exception and the user will see the "Loading" text instead of the editor.

This commit prevents the web component from using an expired access token in `WebComponentLoader` and should thus avoid the exeptions we've been seeing.

However, it's not clear this is actually the correct place to fix the problem, because this ignoring/removal of expired access tokens must already be happening (maybe in `oidc-client`, `redux-oidc`, or associated code?). It might be better if the `WebComponentLoader` only reads the `user` from the redux state (and not also from local storage) and then rely on the component re-rendering when the `user` becomes available in the state.

We know that the `WebComponentLoader` component renders (and therefore makes a request to `editor-api` with no `Authorization` header) before the user is available in state. This is fine for public projects (i.e. those where `user_id` is `null`) but fails for user projects (i.e. those where `user_id` is set). When the user becomes available in state we'd expect the component to re-render and make a request to `editor-api` with the `Authorization` header set correctly. However, this second request doesn't currently happen because of the condition in [this line][2] of `syncProject` in `EditorSlice`, so we'd need to make a change there too.

The main downside of not making the proper fix is that there may be times when the logged-in "state" of different bits of the UI (e.g. the "Save" button, the global nav, etc) might be inconsistent in some circumstances. However, since the proper fix is likely to be a more significant bit of work, it seems sensible to ship this change to prevent the exceptions now and tackle the proper fix separately.

[1]: https://rpf.sentry.io/issues/5135885959/
[2]: https://github.com/RaspberryPiFoundation/editor-ui/blob/e3495a24ff296cf0c3c22910db25d02547c6dfd9/src/redux/EditorSlice.js#L65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+### Changed
+
+### Fixed
+
+- Quick fix for expired access token problem in web component (#1046)
+
 ## [0.25.2] - 2024-13-06
 
 ### Added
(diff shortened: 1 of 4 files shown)


sra405 (approved):
👍 nice! Yea that'll work for now

PR #1051, opened 2024-06-25
Fix local webpack websocket

danhalson:
Fix local webpack websocket error


floehopper (approved):
This looks good to me. I've tried it locally (running within editor-standalone) and it certainly seems to deal with the web socket errors I was seeing in the browser console. 

It's worth noting that I'm seeing some kind of Hot Module Replacement (HMR) errors (JSON parsing error?) in both Chrome & Firefox, but this still feels like a step forward!

floehopper (approved):
I've tried this branch out locally using the web component inside `editor-standalone` and it seems to work nicely - good work! 👍

PR #1062, opened 2024-08-06
Fix Skulpt External Files

sra405:
Fix:

- add external files to document body so it renders outside the shadow DOM

Why:

- Skulpt picks up files by adding them to the DOM (see https://github.com/skulpt/skulpt/issues/854#issuecomment-486135840)
- this doesn't play nice with the shadow DOM and Skulpt doesn't recoginse files like `.txt` and `.csv`
- `createPortal` enables us to pick a common host element (i.e. `body`) and attach the required div's there


conorriches (approved):
s p l e n d i d

PR #1069, opened 2024-08-21
Fix CI to trigger on public forks

sra405:
Currently public forks don't trigger the CI "test" workflow due to it being on push. The workflow "changelog" however does run and that's due to it triggering on "pull_request". Altering "test" to trigger the same way and enable us to approve workflows to run on public forks.


danhalson (approved):
LGTM

PR #1075, opened 2024-09-03
Investigate broken projects site editor web component

sra405:
Currently p5 related projects using the web component in the projects site are broken

### Current findings

- Integrated editor projects in the projects site that utilise p5 error on run (see https://projects.raspberrypi.org/en/projects/editor-target-practice/editor)
- The same editor projects loaded directly in on the editor domain run fine (see https://editor.raspberrypi.org/en/projects/editor-target-practice-starter)
- When break points are added its clear the aspects loaded into p5 via the p5-shim aren't loaded (or in scope at least) when ran in the projects-ui code (i.e. http://localhost:3001/en/projects/editor-rocket-launch/editor) and errors are seen
- Aspects can be initialised in different places or altered slightly so there's less of a dependency on the order they're ran (i.e. removing `Sk.p5.sketch` in place of hardcoding the element ID)
- Moving certain aspects can get a p5 canvas to be rendered in the DOM but fails to load assets or breaks the stop functionality

### Unknowns

- What actually caused this change?
- Why `p5.js` and `p5-shim.js` are being ran in a different order or using difference scopes between projects.raspberrypi.org and editor.raspberrypi.org?
- Both are using the same version web component so why is there a difference?

### DIFF
--- a/src/components/Editor/Runners/PythonRunner/VisualOutputPane.jsx
+++ b/src/components/Editor/Runners/PythonRunner/VisualOutputPane.jsx
@@ -59,11 +59,7 @@ const VisualOutputPane = () => {
       p5Output.current &&
       p5Output.current.innerHTML !== ""
     ) {
-      if (Sk.py5?.stop) {
-        Sk.py5.stop();
-      } else if (Sk.p5?.stop) {
-        Sk.p5.stop();
-      } else if (window.p5?.stop) {
+      if (window.p5?.stop) {
         window.p5.stop();
       } else if (window.py5?.stop) {
         window.py5.stop();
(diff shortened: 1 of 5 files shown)


src/components/Editor/Runners/PythonRunner/VisualOutputPane.jsx
```diff
@@ -56,10 +59,14 @@ const VisualOutputPane = () => {
       p5Output.current &&
       p5Output.current.innerHTML !== ""
     ) {
-      if (Sk.py5.stop) {
+      if (Sk.py5?.stop) {
```
danhalson, line 62: given we're assigning this function to `window.p5` or `window.py5` now, do we need to be checking `Sk.p5` and `Sk.py5` at all anymore?

sra405: nice 👍 good spot, have removed 😄

danhalson (approved):
LGTM

PR #1078, opened 2024-09-19
Set the iframe to sandbox mode to prevent localstorage access

grega:
For context: https://github.com/RaspberryPiFoundation/documentation/pull/112/files

This disables access to localstorage via code executed in the editor:

Screenshot 2024-09-19 at 21 06 50

Results in the following in the console:

Screenshot 2024-09-19 at 20 59 02

***

This works pretty nicely, apart from the rendering of "null preview" in the preview window on the right hand side which will need fixing (it should be "index.html preview").

However, iframe sandbox mode has some potentially unwanted side effects, the most common features disabled are...

* Access to cookies (this cannot be re-enabled, as far as I can tell; re-enabling would also re-enable access to localstorage)
* Form submission is disallowed by default. To allow it, use the allow-forms token.
* The iframe is not allowed to navigate the top-level browsing context (the parent page). To enable top-level navigation, you need allow-top-navigation.
* Opening new browser windows or pop-ups via window.open() is disallowed. You can enable this with allow-popups.
* The iframe is treated as being in a unique origin (different from the parent document). This prevents access to cookies, localstorage, and DOM from the parent unless allow-same-origin is set (which is what we want to achieve of course, at least for localstorage!)
* Pointer lock (which allows capturing the mouse pointer in a specific area) is disallowed unless you explicitly use allow-pointer-lock.
* By default, the iframe cannot make cross-origin requests (e.g., AJAX calls) unless allowed with proper CORS headers and permissions.
* The iframe cannot request fullscreen mode unless allow-fullscreen is used.
* The iframe cannot initiate a file download, even with JavaScript, unless you use allow-downloads.

These might be fine, but will need checking, particularly with the content team. The features above can be [re-enabled using flags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attributes).


grega:
This works pretty nicely, apart from the rendering of "null preview" in the preview window on the right hand side.

However, iframe sandbox mode has some potentially unwanted side effects, the most common features disabled are...

* Form submission is disallowed by default. To allow it, use the allow-forms token.
* The iframe is not allowed to navigate the top-level browsing context (the parent page). To enable top-level navigation, you need allow-top-navigation.
* Opening new browser windows or pop-ups via window.open() is disallowed. You can enable this with allow-popups.
* The iframe is treated as being in a unique origin (different from the parent document). This prevents access to cookies, storage, and DOM from the * parent unless allow-same-origin is set.
* Pointer lock (which allows capturing the mouse pointer in a specific area, useful for games or certain apps) is disallowed unless you explicitly use allow-pointer-lock.
* By default, the iframe cannot make cross-origin requests (e.g., AJAX calls) unless allowed with proper CORS headers and permissions.
* The iframe cannot request fullscreen mode unless allow-fullscreen is used.
* The iframe cannot initiate a file download, even with JavaScript, unless you use allow-downloads.

These might be fine, but will need checking, particularly with the content team. The features above can be re-enabled using flags.

grega:
Have re-enabled sandbox features with no apparent side effect, access to localstorage is still blocked.

Other than further testing, the current issues are still file name not appearing (as mentioned above) and the modal for external link warnings is not being triggered:
Screenshot 2024-09-24 at 10 28 37

grega:
Too restrictive - this blocks all localstorage access which prevents a number of projects from being runnable.

PR #1079, opened 2024-09-25
Disable specific localstorage access at runtime

grega:
Accessing localstorage (eg. `localstorage.getItem('authKey')`) is a potential security risk when running code that has been shared between individuals / is untrusted.

This change overrides the specific localstorage method / key access and provides feedback in the browser console.

Access to localstorage is not entirely blocked as it is used in many projects / resources eg. [Share Your World](https://projects.raspberrypi.org/en/projects/share-your-world) (https://github.com/search?q=repo%3Araspberrypilearning%2Fshare-your-world%20localstorage&type=code)

For more context, see: https://github.com/RaspberryPiFoundation/documentation/pull/112/files

***

The change can be tested by creating a new HTML project in the editor, inputting the following and inspecting the console for the output:

```
<script>
localStorage.setItem("foo", "bar")
console.log(localStorage.getItem("foo"))
localStorage.setItem("authKey", "secret")
console.log(localStorage.getItem("authKey"))
</script>
```

As demonstrated here:

Screenshot 2024-09-25 at 11 47 54

PR #1080, opened 2024-09-25
Broaden scope of localstorage keys disallowed

grega:
Access to all localstorage keys starting with `oidc.` will be disallowed when executing code.

Improves: https://github.com/RaspberryPiFoundation/editor-ui/pull/1079


magdalenajadach (approved):
lgtm

PR #1090, opened 2024-10-03
Web component pyodide support

### LINKED ISSUE #1085: Get PyodideWorker working in the web-component build (locally + deployed)
Currently the PyodideWorker isn't found in the web-component likely due to the `Worker` not doing what it normally does under the hood in a web-component context.

Image

Suggestions:

- look into the worker build process to see if anything needs adding for our `web-component.js` build in `editor-ui`
- utilise an approach investigated in the service worker spike branch https://github.com/RaspberryPiFoundation/editor-ui/pull/904

required as part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/262

loiswells97:
closes #1085


.env.webcomponent.example
```diff
@@ -6,3 +6,4 @@ PUBLIC_URL=http://localhost:3012
 REACT_APP_IDENTITY_URL='http://localhost:3011'
 REACT_APP_API_ENDPOINT='http://localhost:3009'
 ASSETS_URL='http://localhost:3011'
+PUBLIC_URL = "http://localhost:3011"
```
sra405, line 9: 🔧 `PUBLIC_URL` is already defined so needs updating on line 5

loiswells97: Ah, yes, had missed this! Should be sorted now 👍

public/PyodideWorker.js
```diff
@@ -1,7 +1,7 @@
-/* global importScripts, loadPyodide, SharedArrayBuffer, Atomics */
+/* global importScripts, loadPyodide, SharedArrayBuffer, Atomics, pygal, _internal_sense_hat */
 
-import * as pygal from "./pygal.js";
-import * as _internal_sense_hat from "./_internal_sense_hat.js";
+importScripts("http://localhost:3011/_internal_sense_hat.js");
+importScripts("http://localhost:3011/pygal.js");
```
sra405, line 4: 💭 this will only work for local dev. I'm wondering if `importScripts` allows relative imports? Or if we can do something like
```
importScripts(`${PUBLIC_URL}/_internal_sense_hat.js`);
```
it might require some webpack env var parsing though in the webpack config so this is populated on build

public/_internal_sense_hat.js
```diff
@@ -325,33 +348,66 @@ export const gyroRead = () => {
...
 };
 
-export const _read = () => {
+const _read = () => {
   raisePythonError("NotImplementedError", "_read");
 };
+
+_internal_sense_hat = {
+  ..._internal_sense_hat,
+  // config,
```
sra405, line 384: 🪓 
```suggestion
```

public/PyodideWorker.js
sra405: I _think_ this works but sorry for the small comment that threw up a couple of spanners 😆

conorriches:
Thanks for going through this with us Lois, it all makes sense and I'd say looks good to me :)

conorriches (approved):

PR #1106, opened 2024-10-15
Seaborn support

loiswells97:


src/PyodideWorker.js
```diff
@@ -249,19 +254,69 @@ const PyodideWorker = () => {
...
+                bytes_io.seek(0)
+                basthon.kernel.display_event({ "display_type": "matplotlib", "content": bytes_io.read() })
+            plt.show = show_chart
+            `);
           return;
         }
       },
+      after: () => {},
+    },
+    seaborn: {
```
danhalson, line 274: I think we're quickly getting to a point here where sticking these package shim things in separate modules might be preferable

cypress/e2e/spec-wc-pyodide.cy.js
```diff
@@ -100,6 +100,29 @@ describe("Running the code with pyodide", () => {
...
   });
 
+  it("runs a simple seaborn program", () => {
+    runCode("import seaborn as sns\ndata = [50, 30, 100]\nsns.displot(data)");
+    cy.wait(12000);
+    cy.get("editor-wc")
+      .shadow()
+      .find(".pyodiderunner")
+      .find("img")
+      .should("be.visible");
```
danhalson, line 110: would a snapshot be a possibility?

src/PyodideWorker.js
loiswells97: Thanks Dan, Ive created another ticket to look into this 👍

cypress/e2e/spec-wc-pyodide.cy.js
loiswells97: Separate ticket created! 👍

danhalson (approved):
LGTM

PR #1109, opened 2024-10-16
Py-enigma-support

sra405:
Annoyingly py-enigma is relatively old and unmaintained but regularly used in platforms like Ada. The maintainer hasn't included a wheel in the distribution so can't be dynamically installed via micropip. I built one from the `tar.gz` file instead and included alongside or other custom install packages using the command:
```
python -m pip wheel https://files.pythonhosted.org/packages/91/4e/44327ad4a5960de12d86d39e1797f3ab67396a17d82182e8fc1b5ef347e5/py-enigma-0.1.tar.gz
```
where the `tar.gz` file location came from the location in PyPI - https://pypi.org/project/py-enigma/#files


loiswells97 (approved):
Looks great, such a cool library, even if it is somewhat ancient!

PR #1113, opened 2024-10-18
Cache `pyodide` modules

loiswells97:
Refactored the `PyodideWorker` so that `pyodide` is not completely reloaded on each code run, but rather the globals representing user-defined variables and user-imported modules are cleared. This allows the modules that have already been loaded into `pyodide` to be cached, massively reducing the latency on subsequent code runs.


sra405 (approved):
👍

PR #1114, opened 2024-10-21
Fix web component tabbed view

loiswells97:
Small fix to stop web component switching to the pre-set split/tabbed view on every code run

I thought this was related to the `pyodide` switch but it turned out not to be - this is actually an annoying bug that is live on production 😅

### DIFF
--- a/src/components/WebComponentProject/WebComponentProject.jsx
+++ b/src/components/WebComponentProject/WebComponentProject.jsx
@@ -52,9 +52,11 @@ const WebComponentProject = ({
   const [codeHasRun, setCodeHasRun] = useState(codeHasBeenRun);
   const dispatch = useDispatch();
 
-  dispatch(setIsSplitView(outputSplitView));
-  dispatch(setWebComponent(true));
-  dispatch(setIsOutputOnly(outputOnly));
+  useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/components/WebComponentProject/WebComponentProject.jsx
```diff
@@ -52,9 +52,11 @@ const WebComponentProject = ({
...
   const dispatch = useDispatch();
 
-  dispatch(setIsSplitView(outputSplitView));
-  dispatch(setWebComponent(true));
-  dispatch(setIsOutputOnly(outputOnly));
+  useEffect(() => {
+    dispatch(setIsSplitView(outputSplitView));
+    dispatch(setWebComponent(true));
+    dispatch(setIsOutputOnly(outputOnly));
+  }, [outputSplitView, outputOnly, dispatch]);
```
magdalenajadach, line 59: I wonder if the dispatch could be skipped sine it's a stable reference

loiswells97: We need these dispatches to put the data from the web component attributes into state. But you're right, once it's rendered the first time, these values shouldn't change - the dependencies in the array are just there to keep the linter happy (annoyingly).

magdalenajadach (approved):
lgtm

PR #1122, opened 2024-10-22
v0.28.3

loiswells97:


CHANGELOG.md
```diff
@@ -928,7 +935,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Events in Web Component indicating whether Mission Zero criteria have been met (#113)
 
-[unreleased]: https://github.com/RaspberryPiFoundation/editor-ui/compare/v0.28.2...HEAD
+[unreleased]: https://github.com/RaspberryPiFoundation/editor-ui/compare/v0.28.3...HEAD
+[0.28.3]: https://github.com/RaspberryPiFoundation/editor-ui/releases/tag/v0.28.2
```
sra405, line 939: copy & 🍝 
```suggestion
[0.28.3]: https://github.com/RaspberryPiFoundation/editor-ui/releases/tag/v0.28.3
```

loiswells97: Oh no! 🤦‍♀️

src/PyodideWorker.js
```diff
@@ -260,7 +260,12 @@ const PyodideWorker = () => {
           return;
         }
       },
-      after: () => {},
+      after: () => {
+        pyodide.runPython(`
+        import matplotlib.pyplot as plt
+        plt.clf()
+        `);
```
sra405, line 267: ❓ it makes sense but I also thought anything without a `_` got wiped? I'm just surprised `pyplot` made it through 🤔 weird

loiswells97: Hmmm that's true, I'm not sure - I'd assumed it was because matplotlib isn't getting reloaded, and somehow it remembered what the graph was meant to be... 🤷‍♀️

loiswells97: I can dig deeper though if you like... 🤔

sra405: na its ok, I think we might just need to keep this sort of thing in mind, no everything is wiped and it can lead to weird behaviour is all

sra405 (approved):
👍 nice! Love a good `plt.clear`

src/PyodideWorker.js
loiswells97: Okay, I just had a look and it appears that `plt` itself isn't making it through if you run again without defining it... So it must just remember what was in the plot another way and then be putting it back when they try to make a new graph. Seems odd though, like you say

PR #1124, opened 2024-10-23
set REACT_APP _API_ENDPOINT and _AUTHENTICATION_URL as props

conorriches:
## Issue
Closes https://github.com/RaspberryPiFoundation/learner-experience-integration-tests/issues/5

## Description

* Changed how both utils function:
  * `src/utils/apiCallHandler.js`
  * `src/utils/userManager.js`
* I've considered how best to adapt these helpers - I could have gone with classes or functions, and while I considered changing these to be classes, this would require more verbose invocation throughout the codebase which ended up making code harder to read, whereas the function-based approach means most calls can be left as-is.
* `public/index.html` added which provides a handy link to the web-component.html rather than just a directory listing
* Removed a load of code that isn't used any more


src/utils/userManager.js
```diff
@@ -5,13 +5,13 @@ const host = `${window.location.protocol}//${window.location.hostname}${
   window.location.port ? `:${window.location.port}` : ""
 }`;
 
-const userManagerConfig = {
+const userManagerConfig = ({ reactAppAuthenticationUrl }) => ({
```
sra405, line 8: ❓ I'm struggling to see where `reactAppAuthenticationUrl` is set 🤔 I'm sure its just me but I can't see it 😅 or do we not need it?

src/redux/EditorSlice.test.js
```diff
@@ -16,14 +7,25 @@ import reducer, {
...
+const mockReadProject = jest.fn();
+const mockCreateOrUpdateProject = jest.fn();
+
+jest.mock("../utils/apiCallHandler", () => () => ({
+  createRemix: jest.fn(mockCreateRemix),
+  deleteProject: jest.fn(mockDeleteProject),
+  loadAssets: jest.fn(mockLoadAssets),
+  readProject: jest.fn(mockReadProject),
+  createOrUpdateProject: jest.fn(mockCreateOrUpdateProject),
+}));
```
sra405, line 28: ❤️ this actually makes testing below read nicer imo

src/utils/userManager.js
conorriches: Yeah I had the same realisation when doing this 😅 
It's actually not yet used - only in `src/app/store.js` where I've left a TODO:

```
// TODO - not used but keeping this in preparation for using
// src/components/Editor/ImageUploadButton/ImageUploadButton.jsx
```

src/containers/WebComponentLoader.jsx
```diff
@@ -31,24 +32,25 @@ const WebComponentLoader = (props) => {
...
-    showSavePrompt = false,
+    identifier,
+    instructions,
     loadRemixDisabled = false,
-    readOnly = false,
     outputOnly = false,
+    outputPanels = ["text", "visual"],
     outputSplitView = false,
+    projectNameEditable = false,
+    reactAppApiEndpoint = process.env.REACT_APP_API_ENDPOINT,
```
sra405, line 45: 💪 nice!! So given this is a default fallback do you reckon its worth keeping this env var in the `.env.example` and CI/CD so it'll work as is but can be overridden in the `editor-wc` props?

src/utils/userManager.js
sra405: that's mad that's the only use of that variable 😅

src/containers/WebComponentLoader.jsx
```diff
@@ -135,6 +138,10 @@ const WebComponentLoader = (props) => {
     saveTriggered,
   });
 
+  useEffect(() => {
+    dispatch(setReactAppApiEndpoint(reactAppApiEndpoint));
+  }, [reactAppApiEndpoint, dispatch]);
```
sra405, line 143: 🔧 definitely could do with a test for this in the `WebComponentLoader.test.js` 👍

src/containers/WebComponentLoader.jsx
conorriches: Sure! done in [727da9c](https://github.com/RaspberryPiFoundation/editor-ui/pull/1124/commits/727da9c16abb1638406c76afbf58544dab6ddbe8) :)

src/web-component.html
```diff
@@ -43,6 +43,7 @@
       // sidebar
       webComp.setAttribute("with_projectbar", "true");
       webComp.setAttribute("with_sidebar", "true");
+      webComp.setAttribute("react_app_api_endpoint", "https://test-editor-api.raspberrypi.org");
```
sra405, line 46: 🔧  I'd be tempted removing this from the test page and letting it get set in CI still
```suggestion
```

sra405 (approved):
💪 this is awesome!

PR #1132, opened 2024-11-06
EditorInput - remove auto scroll on content change

sra405:
Fixes - https://raspberrypifoundation.slack.com/archives/C022FDFRM7B/p1730800544670889

### DIFF
--- a/src/components/Editor/EditorInput/EditorInput.jsx
+++ b/src/components/Editor/EditorInput/EditorInput.jsx
@@ -1,4 +1,4 @@
-import React, { createRef, useEffect, useRef, useState } from "react";
+import React, { createRef, useEffect, useRef, useState, useMemo } from "react";
 import { DragDropContext } from "@hello-pangea/dnd";
 import { useDispatch, useSelector } from "react-redux";
 import { TabPanel, Tabs } from "react-tabs";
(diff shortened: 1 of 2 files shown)


src/components/Editor/EditorInput/EditorInput.jsx
```diff
@@ -95,19 +97,26 @@ const EditorInput = () => {
   }, [project?.components]);
 
   useEffect(() => {
-    if (!project?.components) return;
+    const newFileNames = project.components.map(
+      (file) => `${file.name}.${file.extension}`,
+    );
+    if (newFileNames.join() !== fileNames?.join()) {
```
conorriches, line 103: Just checking - things won't be in a different order right?

conorriches (commented):
Nice! Quick question that's all

src/components/Editor/EditorInput/EditorInput.jsx
sra405: nice thinking and challenge. Just checking but it shouldn't matter. The only way we used to be able to alter the file order is through drag/drop. But the features currently not working and even if it was you'd probably want to scroll then anyway. This current approach scrolls for opened/reopened files but doesn't scroll for anything that doesn't change this array (i.e. `project.component.content` changes that currently invoke this on type)

conorriches: Roger! Makes sense

conorriches (approved):
Very nice!

PR #1145, opened 2024-11-27
Check loadRemixDisabled

sra405:
Initial attempt at a fix for auto loading remixes when we explicitly don't want to to avoid editor projects remixing, loading the remix then causing a url changed based off the load rather than remix response


loiswells97 (commented):
Looks good - it's kicking off about the linting though 😅

sra405:
> Looks good - it's kicking off about the linting though 😅

haha hadn't spotted that!! Classic

loiswells97 (approved):
Looks great 👍

PR #1147, opened 2024-11-29
Fix p5 run errors

sra405:
Reported that when running p5 code more than once the runner breaks and errors.

This remove setting visualOutput to false and uses initial state and rerenders to set the correct visual output behaviour.

This is still dynamic and works well when adding code that requires visualisation as well as always showing for the sense hat model


loiswells97 (approved):
Approved ✅

PR #1150, opened 2024-12-05
Fix dark mode selection highlighting

loiswells97:
- Fixed the selector specificity for selected text in the editor
- Toned down the match highlighting

### Before
Screenshot 2024-12-05 at 14 31 10

### After
Screenshot 2024-12-05 at 14 19 57


maxelkins:
Thanks for sharing :)

From my eye the match highlighting looks a lot better, getting pefect contrast with all the different colours will be a tricky job. I think lets get this in and revisit the code mirror theme as a whole later.

PR #1152, opened 2024-12-12
190 login to save only logs in

danhalson:
Closes https://github.com/RaspberryPiFoundation/editor-standalone/issues/190

### DIFF
--- a/src/containers/WebComponentLoader.test.js
+++ b/src/containers/WebComponentLoader.test.js
@@ -315,7 +315,7 @@ describe("When no user is in state", () => {
               instructions={instructions}
               authKey={authKey}
               theme="light"
-              useCache={true}
+              loadCache={true}
             />
           </CookiesProvider>
         </Provider>,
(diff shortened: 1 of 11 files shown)


src/containers/WebComponentLoader.test.js
```diff
@@ -315,6 +315,7 @@ describe("When no user is in state", () => {
               instructions={instructions}
               authKey={authKey}
               theme="light"
+              useCache={true}
```
sra405, line 318: ❓ I don't think this is doing anything. The prop is `loadCache` and I _think_ the test below are just picking up the default

src/web-component.js
```diff
@@ -86,6 +87,7 @@ class WebComponent extends HTMLElement {
         "use_editor_styles",
         "with_projectbar",
         "with_sidebar",
+        "use_cache",
```
sra405, line 90: 🔧 is this not feeding into `loadCache`? So should be `load_cache`?

sra405 (approved):
👍 nice!

PR #1158, opened 2025-01-02
Fix skulpt support for output tabs attribute

loiswells97:


src/components/Editor/Runners/PythonRunner/PythonRunner.jsx
```diff
@@ -93,7 +93,10 @@ const PythonRunner = () => {
   return (
     <>
       <PyodideRunner active={activeRunner === "pyodide"} />
-      <SkulptRunner active={activeRunner === "skulpt"} />
+      <SkulptRunner
```
danhalson, line 96: Wonder if it's worth adding a simple test to check the outputPanels prop is getting added? just for a bit of regression coverage (and to help avoid it getting missed in the future)

loiswells97: Thanks Dan, I've added a couple of tests to check this (although in a bit of a roundabout way, as I think React Testing Library doesn't let you test which props are on components, just what's in the DOM annoyingly)

danhalson (approved):
LGTM

PR #1166, opened 2025-01-16
Feat/editable instructions

loiswells97:


conorriches (approved):
all code to this point has had a PR and reviews, having taking a final skim through this looks good to me

PR #1169, opened 2025-01-20
Fix cached instructions

loiswells97:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/392


danhalson (approved):
Looks great, might just need to be careful that there wasn't any unintentional reliance on the state bleeding between mounts, pretty sure there isn't since I've looked into this logic too, but just worth keeping an eye on it

PR #1171, opened 2025-01-22
Writing binary files

loiswells97:
At the moment, the batching means that when an image is created, the main thread interprets that as separate images and creates a bunch of invalid files with the same name. To stop this happening, we probably need to modify the response on the frontend to ensure that the batched writes all hit the same file and append to rather than overwrite the existing content. The question we will have is when do we want existing files to be overwritten and when do we want them to be appended to and how can we tell? For example, if it's another batch of binary from the same request then we might want it to append, whereas if it's another request from the user, we might want it to overwrite. Maybe we need some metadata to give a unique ID to the request?

Question: How do we want to deal with image files created by a user who is not logged in? Do we want to let them save the image to the bucket anyway? Or should it be saved in localStorage somehow? An MVP option could be to tell them they have to log in to add images to the project, although not ideal.

Also, we may need some changes on the API side to stop writing to a filename that already exists creating multiple files with the same name in the same project. Can we have an update image endpoint as well as a create image endpoint? And if so, would we overwrite the original image in the bucket? In other words, how would this work in practice?

Work out why updating an image only works after the page has been reloaded, otherwise it creates a second image with the same name.

### DIFF
--- a/src/hooks/useProjectPersistence.js
+++ b/src/hooks/useProjectPersistence.js
@@ -7,7 +7,6 @@ import {
   syncProject,
 } from "../redux/EditorSlice";
 import { showLoginPrompt, showSavePrompt } from "../utils/Notifications";
-import { base64ToUint8Array } from "../utils/base64Helpers";
 
 const COMBINED_FILE_SIZE_SOFT_LIMIT = 1000000;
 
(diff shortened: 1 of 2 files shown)


loiswells97:
## Where things are at

- There's a merge conflict with the main branch since the file write stuff was merged - needs a rebase to fix it
- It works fine for ordinary binary file write
- Imageio imwrite appears to call the binary write function multiple times (16 times?) per image write, and there is a race condition or variable not updating in the main thread which means that these requests are all being fired off before the first one is completed, resulting in multiple files of the same name being created, none of which are valid image files.

PR #1176, opened 2025-01-24
Add remove instructions button modal

jamiebenstead:
Work for [this](https://github.com/orgs/RaspberryPiFoundation/projects/51?pane=issue&itemId=92785863&issue=RaspberryPiFoundation%7Cdigital-editor-issues%7C373) ticket.

### DIFF
--- a/src/utils/i18n.js
+++ b/src/utils/i18n.js
@@ -220,6 +220,8 @@ i18n
             nextStep: "Next step",
             previousStep: "Previous step",
             projectSteps: "Project steps",
+            edit: "Edit",
+            view: "View",
             removeInstructionsModal: {
               removeInstructions: "Remove instructions",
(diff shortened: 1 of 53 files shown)


conorriches (commented):
This is excellent, great small refactors in here too. 

I'm happy to approve, but I think cypress may need fixing first

loiswells97 (approved):

src/utils/i18n.js
```diff
@@ -208,6 +208,7 @@ i18n
           instructionsPanel: {
             emptyState: {
               addInstructions: "Add instructions",
+              removeInstructions: "Remove instructions",
```
loiswells97, line 211: My only question is whether this should sit outside of the emptyState layer as it's in the view where there are instructions?

PR #1178, opened 2025-01-28
Try pinning previous ubuntu version to avoid cloudflare issue

danhalson:
Test deployment to try out a potential workaround for cloudflare issue:

```
Update - We are continuing to work on a fix for native CRC32 checksum handling in R2 APIs. If you run into an issue, please see the documentation for workarounds based on the SDK that you are using: https://developers.cloudflare.com/r2/examples/aws/
Jan 22, 2025 - 16:33 UTC
Update - We are continuing to work on a fix for this issue.
Jan 17, 2025 - 11:23 UTC
Identified - AWS recently updated their SDKs to enable CRC32 checksums on multiple object operations by default. 
R2 does not currently support CRC32 checksums, and the default configurations will return header related errors such as Header 'x-amz-checksum-algorithm' with value 'CRC32' not implemented. 

Impacted users can either pin AWS SDKs to a prior version or modify the configuration to restore the prior default behavior of not checking checksums on upload. 

For more details, see the examples section of the R2 Docs for the relevant SDK: 
https://developers.cloudflare.com/r2/examples/aws/
Jan 17, 2025 - 03:20 UTC
```

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   changelog-updated:
-    runs-on: ubuntu-22.03
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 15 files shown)


conorriches (approved):
hopefulyl the changes to whitespace won't do anything, let's see!

PR #1182, opened 2025-01-30
fix dark mode colours

conorriches:
image


loiswells97 (approved):
Looks sooo much better, thanks for jumping on this and fixing so quick!

PR #1191, opened 2025-02-04
Fixing instructions state when set to an empty string

loiswells97:
This is fixing a bug where the user deletes everything in the instructions box and it suddenly switches back to the empty instructions state. This was caused by the fact that an empty string is `falsy` in Javascript, so when the instructions content was deleted we would get `projectInstructions = ""` which resulted in the instructions being deleted.


danhalson (approved):
LGTM

PR #1196, opened 2025-02-20
Check instructions length before setting option value

jamiebenstead:
[closes](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/443)

I found that `const option = instructions ? "instructions" : "file";` was always being set to true because it was receiving an empty array

### DIFF
--- a/src/components/Menus/Sidebar/SidebarBar.jsx
+++ b/src/components/Menus/Sidebar/SidebarBar.jsx
@@ -25,10 +25,7 @@ const SidebarBar = (props) => {
   const isMobile = useMediaQuery({ query: MOBILE_MEDIA_QUERY });
 
   const expandPopOut = () => {
-    let option = "file";
-    if (instructions.length > 0) {
-      option = "instructions";
-    }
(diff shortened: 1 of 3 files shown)


conorriches (approved):
great spot!

src/components/Menus/Sidebar/SidebarBar.jsx
```diff
@@ -25,7 +25,10 @@ const SidebarBar = (props) => {
   const isMobile = useMediaQuery({ query: MOBILE_MEDIA_QUERY });
 
   const expandPopOut = () => {
-    const option = instructions ? "instructions" : "file";
+    let option = "file";
+    if (instructions.length > 0) {
+      option = "instructions";
+    }
```
danhalson, line 31: you could make this a one liner again with `const option = instructions.length > 0 ? "instructions" : "file"` but happy with it either way 😄

danhalson (approved):
Approving as it's all fine, but added a suggestion which you can take or leave!

PR #1200, opened 2025-03-24
Fixing the append file write mode

loiswells97:
## What's Changed?

- Fixed the mode on the `pyodide` side - it was always using `w` rather than the mode provided in the code
- Fixed the file write append mode on the frontend - it was incorrectly adding a newline character before anything was appended.


src/PyodideWorker.js
danhalson, line 1: are there any tests around this?

PR #1203, opened 2025-04-09
Fix turtle canvas bug

loiswells97:
This PR is very much WIP
Although these changes fix the original bug, they introduce an issue with the `p5` canvas that needs investigating before this can go out.

Steps to reproduce:
1. Run a basic `p5` program
2. Delete that code and run `print("hello world")`
3. Run the basic `p5` program again

Expected behaviour: Output from the `p5` program is displayed in the visual output tab
Actual behaviour: The visual output tab is rendered, but no output is visible

closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/137

PR #1207, opened 2025-04-15
v0.30.0

loiswells97:
### Added
- `editor-projectLoadFailed` custom event that fires when a project completely fails to load (#1201)
- Added runnerBeingLoaded state to prevent race condition overwrites (#1205)

### Fixed
- Bugs in append mode for writing to files in python (#1200)
- `turtle` bug that did not display output on first code run (#1203)


adrian-rpf (approved):
Step up

PR #1219, opened 2025-06-17
fix: change tab `block-size` to `min-block-size` to allow height to grow

maxelkins:
Closes: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/690

## Summary

**Issue related:**
- Change tab `block-size` to `min-block-size` to allow height to grow
- Also change to correct custom property to ensure minimum target size

**Fixing failing tests:**
- The Skulpt runner implementation was hiding tab/elements by setting block-size (height) to 0. As a result when using `min-block-size` this was causing cypress visiblity test to fail.
- It was previously passing since cypress considers height 0 as not visible.
- This implementation has now been changed to set `display: none` to properly remove the element
- There are 2 other potential changes that could be made here
  1.  Conditially render the elements instead - as the [PyodideRunner](https://github.com/RaspberryPiFoundation/editor-ui/blob/main/src/components/Editor/Runners/PythonRunner/PyodideRunner/PyodideRunner.jsx#L440) does
  2.   Change the cypress test to check for [existence](https://docs.cypress.io/api/commands/its#Existence) instead
- I've opted not to change the above to keep this PR smaller

## Demo
Screenshot 2025-06-17 at 16 32 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Styling issue preventing scrolling in the sidebar (#1216)
 - Styling issue on status bar on mobile (#1217)
 - Change tab `block-size` to `min-block-size` to allow for height increase when text wraps (#1219)
-- Changed `SkulptRunner.jsx` implementation of hiding elements to used `display: none` rather than `block-size: 0` (#1219)
+- Changed `SkulptRunner.jsx` implementation of hiding elements to use `display: none` rather than `block-size: 0` (#1219)
 
 ## [0.30.1] - 2025-06-09
 


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Styling issue on sidebar on mobile (#1194)
 - Styling issue preventing scrolling in the sidebar (#1216)
+- Change tab `block-size` to `min-block-size` to allow for height increase when text wraps (#1219)
```
danhalson, line 23: shouldn't this be in the unreleased section?

maxelkins: woops!

maxelkins: Fixed. Also moved Jamies one which was in the wrong place

danhalson (approved):
Great work!

PR #1223, opened 2025-06-27
Add translations to times and numbers

jamiebenstead:
Editor-UI part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/588

### DIFF
--- a/src/utils/formatRelativeTime.js
+++ b/src/utils/formatRelativeTime.js
@@ -11,9 +11,9 @@ export const formatRelativeTime = (
   lastSavedTime,
   now,
   languageCode,
-  fallbackLgn,
+  fallbackLng,
 ) => {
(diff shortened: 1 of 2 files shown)


src/utils/formatRelativeTime.js
```diff
@@ -0,0 +1,16 @@
+import { intlFormatDistance } from "date-fns";
+
+const localeMap = {
```
danhalson, line 3: I'm lacking a bit of context here, but are we redefining this list, rather than using a shared source of truth?

loiswells97: I think the answer is that this map is for ensuring that our locales match up with the ones from the library

danhalson: could we move this into a central location, maybe a separate util called `localeMappings.js` or something, and make sure nowhere else is redefining it?

PR #1229, opened 2025-07-16
Upgrading to design system react v2.4.0

loiswells97:
## What's Changed

- `design-system-react` has been upgraded to version `2.4` - this removes the problematic code that was being injected into the head. This caused a build failure because it was attempting to load the font from the wrong path when the design system was used in a web component.
- Removed import of `design-system-core` as this is now packaged with `design-system-react`
- Material Symbols is now imported into the `ExternalStyles` and injected into the host page, as there seems not to be a way to get the symbols to load up from within the web component. Although this is unideal, we are looking to move away from Material Symbols in the future.

closes #https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
+- Issue where clicking the tick icon would not save a filename when editing (#1237)
 - Fixed loading of Material Symbols font (#1229)
 
 ## [0.30.2] - 2025-07-22
--- a/src/components/ProjectName/ProjectName.jsx
+++ b/src/components/ProjectName/ProjectName.jsx
@@ -74,29 +74,20 @@ const ProjectName = ({
     }
   };
 
+  const handleOnBlur = (event) => {
+    // If the blur event is triggered by clicking the tick button we want to make sure updateName is called
+    if (tickButton?.current?.contains(event.relatedTarget)) {
+      event.preventDefault();
+      return;
+    }
+    resetName(event);
+  };
+
   useEffect(() => {
     if (isEditing) {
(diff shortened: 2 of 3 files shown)


maxelkins:
I think this solves the issue:

In `ExternalStyles.scss`

```scss
@use "../../../node_modules/react-tabs/style/react-tabs.css";
@use "../../../node_modules/react-toggle/style.css";
@use "../../../node_modules/prismjs/plugins/line-numbers/prism-line-numbers.css";
@use "../../../node_modules/prismjs/plugins/line-highlight/prism-line-highlight.css";

// Remove only import alert styles
// @use "../../../node_modules/@raspberrypifoundation/design-system-core/scss/components/alert.scss";

// Remove importing icon font via the package. Should probably also remove the package?
// @use "../../../node_modules/material-symbols/sharp.scss";

// Imports all design system styles as per https://github.com/RaspberryPiFoundation/design-system-react/blob/main/UPDATING.md
@forward "@raspberrypifoundation/design-system-react/scss/design-system-core";

// Load the icon fonts required
@import url("https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined");
@import url("https://fonts.googleapis.com/css2?family=Material+Symbols+Sharp");

```

See https://github.com/RaspberryPiFoundation/design-system-react/blob/main/UPDATING.md for more detail on the way we changed including code in design system react v2+

DanielBrierton:
Thanks @maxelkins, that was really useful. Have it working now. I was able to just use `@import` in `ExternalStyles.scss` though. Got a link to something that says that shouldn't work, or am I misunderstanding your comment? 😅

I only added Material Symbols Sharp, since that's all we actually use, and I think we probably don't want to expand our usage of Material Symbols at this point. 😶

PR #1233, opened 2025-07-22
Add styling to allow for scrollbars to be themed in Chrome

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/776


src/components/Editor/EditorPanel/EditorPanel.jsx
```diff
@@ -164,7 +164,7 @@ const EditorPanel = ({ extension = "html", fileName = "index" }) => {
   }, [file, cascadeUpdate, editorViewRef]);
 
   return (
-    <>
+    <div className={isDarkMode ? "dark" : "light"}>
```
loiswells97, line 167: I think we already have class names more globally that could be used for some of this? Unless we got rid of them 🤔

loiswells97 (approved):
Looks good ✅

PR #1237, opened 2025-07-28
Fix unable to rename a file

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/797

### DIFF
--- a/src/components/ProjectName/ProjectName.jsx
+++ b/src/components/ProjectName/ProjectName.jsx
@@ -76,7 +76,7 @@ const ProjectName = ({
 
   const handleOnBlur = (event) => {
     // If the blur event is triggered by clicking the tick button we want to make sure updateName is called
-    if (event.relatedTarget?.className?.includes("project-name__button")) {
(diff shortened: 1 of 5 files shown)


src/components/ProjectName/ProjectName.jsx
```diff
@@ -74,30 +74,21 @@ const ProjectName = ({
...
   };
 
+  const handleOnBlur = (event) => {
+    // If the blur event is triggered by clicking the tick button we want to make sure the target is clicked
+    if (event.relatedTarget?.className?.includes("project-name__button")) {
+      const target = event.relatedTarget ? event.relatedTarget : null;
+      target.click();
+    }
+    resetName(event);
+  };
```
loiswells97, line 84: Can we not just not reset the name if the click is over the button and let the button handle the click event? I think we're trying to avoid having fake clicks firing off and things 🤔

loiswells97 (approved):
Looks great, let's get it out there!

PR #1239, opened 2025-08-01
Set selectionMatch to be transparent

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/721


loiswells97 (approved):
Nice, I think this is the simplest solution that allows us to turn this back on if we ever need to 👍

PR #1240, opened 2025-08-07
Update to v0.30.3

jamiebenstead:
### Fixed

- Issue where clicking the tick icon would not save a filename when editing (#1237)

### Changed

- Set selectionMatch to be transpartent so not all instances of a word are highlighted when selected (#1239)


danhalson (approved):
LGTM

PR #1244, opened 2025-09-18
Add locale attribute instead of detecting language

loiswells97:
Switches to determining the language from an attribute rather than detecting through `i81n`. This maintains independence from the setup of the host page whilst allowing the web component to be consistent with the host in its choice of language.

closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/883


adrian-rpf (commented):
Good code and changes make sense. 
⚠️ Requires update to the `readme.md` for new attribute

PR #1251, opened 2025-10-06
Plugin system for sidebar panels

loiswells97:


src/components/Editor/Project/Project.jsx
```diff
@@ -22,6 +22,7 @@ const Project = (props) => {
     withProjectbar = true,
     withSidebar = true,
     sidebarOptions = [],
+    plugins = [],
```
adrian-rpf, line 25: wondering if we should name this sidebarPlugins to limit scope/.

src/components/Menus/Sidebar/Sidebar.jsx
```diff
@@ -79,6 +80,22 @@ const Sidebar = ({ options = [] }) => {
     },
   ].filter((option) => options.includes(option.name));
 
+  let pluginMenuOptions = plugins.map((plugin) => {
```
adrian-rpf, line 83: naming?

adrian-rpf (commented):
I really like this code. Really nice and clear. 

1. Should we consider the naming of some of the variables. Limiting or making it clear that that it's for use for the sidebar.

2. How will we document how to use this plugin system? ReadMe, Example Plugin Script?

adrian-rpf (approved):
Really nice change and good plan. 
We have discussed this needs documenting/creating an example at a later point.

PR #1271, opened 2025-11-06
Refactor SidebarPanel to pass in buttons as elements

jamiebenstead:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/991

### DIFF
--- a/src/components/Menus/Sidebar/SidebarPanel.jsx
+++ b/src/components/Menus/Sidebar/SidebarPanel.jsx
@@ -16,14 +16,12 @@ const SidebarPanel = (props) => {
   } = props;
   const isMobile = useMediaQuery({ query: MOBILE_MEDIA_QUERY });
 
-  const buttonNodes = React.Children.toArray(buttons);
-
   const panelContent = (
     <>
       <div className="sidebar__panel-header">
         <h2 className="sidebar__panel-heading">{heading}</h2>
-        {buttonNodes.length > 0 && (
-          <div className="sidebar__panel-buttons">{buttonNodes}</div>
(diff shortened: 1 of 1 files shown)


src/components/Menus/Sidebar/InstructionsPanel/InstructionsPanel.jsx
```diff
@@ -147,25 +147,25 @@ const InstructionsPanel = () => {
...
+                  className="btn--primary"
+                  icon="add"
+                  text={t("instructionsPanel.emptyState.addInstructions")}
+                  onClick={addInstructions}
+                  fill={true}
+                  textAlways={true}
+                  small={true}
+                />,
               ]
           : null
```
loiswells97: I think it should be empty if there are no buttons rather than null (although I suspect the default on the component is handling this

src/components/Menus/Sidebar/ProjectsPanel/ProjectsPanel.jsx
```diff
@@ -38,14 +39,15 @@ const ProjectsPanel = () => {
...
-        },
+        <DesignSystemButton
+          key="your-projects"
+          className="btn--primary projects-panel__your-projects-button"
+          onClick={navigateToProjectsPage}
+          text={t("projectsPanel.yourProjectsButton")}
+          textAlways
+        />,
       ]
-    : [];
```
loiswells97: Same with this, does it need to change to null?

loiswells97 (approved):
Nice one, thanks for putting up with my prevaricating over the best approach 😅

PR #1275, opened 2025-11-13
1004: Prevent access to the session across all scopes, and tighten localStorage access scope

danhalson:
- Prevent access to the session across all scopes
- Tighten localStorage access scope


src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -306,55 +306,90 @@ function HtmlRunner() {
...
+              if (!host) return;
+              try {
+                Object.defineProperty(host, "sessionStorage", {
+                  configurable: false,
+                  enumerable: false,
+                  get: () => stub,
+                  set: () => undefined,
+                });
+              } catch (_) {
+                /* ignore if non-configurable */
```
grega: Same as above

PR #1308, opened 2026-02-05
Projects site integrated editor improvements

jamdelion:
- **Hide inline-end border for selected tab**
- **Add button border radius property**

This PR adds some variables that can be used in the projects site to customise the style of the integrated editor projects. It also adds a pseudo-element that can be toggled with a variable to affect the styling of the sidebar tabs.

The designs for this are here: https://www.figma.com/design/2XzRDyXtthyzeNMZtdChTQ/Projects-site---Designs?node-id=10785-10788&m=dev

See corresponding projects site changes here: https://github.com/RaspberryPiFoundation/projects-ui/pull/3611


jamdelion:
> Changes look good,
> 
> Because the level of customisation, it seems possible that a future editor release might break something by introducing something new that is unstyled or by breaking one of the overrides.
> 
> Is this something you will check for an manage when you upgrade the editor-ui component? I was just wondering if it's worth doing anything to easier to check as we're working on the code editor such having an example in the project that uses the overrides.

Thanks @zetter-rpf, that sounds sensible. Would you recommend using storybook for this, or something else?

PR #1319, opened 2026-02-11
Improve the look of editor preview

zetter-rpf:
Use the with_editor_styles setting by default.

The editor looks broken without these styles - items with poor contrast and elements without the correct padding.

In the longer term we may want to have a simpler 'default' styling that looks good.

As before you can still configure use_editor_styles it by using a URL param if you want to see the editor without styles.

|before|after|
|---|---|
|localhost_3011_web-component html(iPad Pro) (2)|localhost_3011_web-component html(iPad Pro)|


adrian-rpf (approved):
Simple change, big improvement

PR #1333, opened 2026-02-18
Remove material symbols

zetter-rpf:
I did some investigation into our bundle size. Material symbols was taking up 80% even though we don't use many symbols (only the alert symbol for one error message)

### Bundle size:

Screenshot 2026-02-18 at 15 53 28

Not loading material symbols will make the code editor load faster and perform better in slower devices.

We're likely to want to use symbols in the future, especially since other components in the design library use material symbols. Then we have a few options:

+ Instead of using material symbols, use our own icons
+ Include just the material symbol font we're using from the material-symbols package
+ Require the host page to load the material symbols font. The web component would just need to define styles on '.material-symbols-sharp' as the material-symbols library does.

We could apply one of these now, but I think it would be helpful to have more example of symbols we want to use before we do which is why my preference is to remove them for now.

### Visual changes
The only change is to remove the symbol used for an alert that I expect a very small proportion of our users see. Before:

Screenshot 2026-02-18 at 15 25 06

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "jszip": "^3.10.1",
     "jszip-utils": "^0.1.0",
     "marked": "^15.0.6",
-    "material-symbols": "^0.27.0",
     "mime-types": "^2.1.35",
     "node-html-parser": "^6.1.5",
     "oidc-client": "^1.11.5",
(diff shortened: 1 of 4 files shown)


maxelkins:
I think ideally we'd use the icon, since it just keeps the design of that component consistent with other places we use it. But its hard to argue shipping the font file for that 1 icon, so I understand the choice to remove it 👍

Maybe we just `display: none` `.rpf-alert__icon` so it looks like:
Screenshot 2026-02-18 at 16 27 04

PR #1342, opened 2026-02-25
Only retrieve cached projects for the current locale

cocomarine:
### Locale switching fix
- Before: caching wasn't working properly, hence the editor-standalone not behaving consistently and the CCP broken with locale switching (see https://www.figma.com/board/21lOTDiDkQ0cdUNF2fpnZh/Editor-behaviour-on-locale-change?node-id=0-1&p=f&t=apL6cwEJUa2Vod7Q-0)
- Added a locale match check before loading a cached project in `useProject` hook
  - **This is a temporary fix as it effectively disables caching.**
  - Both editor-standalone and projects-ui behave in the same way after this change. 
- Updated existing caching-related tests in `useProject.test`
### Initial load problem fix
- Before: when an editor page with locales other than en was first loaded in CCP, it rendered en page rather than the locale page (editor-standalone was fine)
- Updated to make sure initial attributes like locale are passed to React by reading them from the DOM in `reactProps()` of `web-component.js` 
  - This is so that the first render uses the host's value even when `attributeChangedCallback` hasn't been run yet

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,7 +19,7 @@ yarn start
 ```
 
 ## Development Workflow
-- Node.js: use v18.x locally (as pinned in `.tool-versions`). 
+- Node.js: use the version pinned in `.tool-versions`.
 - CI currently runs on Node 16, so avoid using Node APIs or syntax that are not supported in Node 16 until CI is updated or aligned.
 - Yarn 4 is required (`packageManager` in `package.json`). If you don't have the right Yarn version available, run `corepack enable`. `npm install` can fail - use `yarn install` instead.
 - Dev server: `yarn start` (webpack dev server on `http://localhost:3011`).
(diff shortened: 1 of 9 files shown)


src/hooks/useProject.js
```diff
@@ -49,7 +51,13 @@ export const useProject = ({
       const is_cached_unsaved_project =
         !projectIdentifier && cachedProject && !initialProject;
 
-      if (loadCache && (is_cached_saved_project || is_cached_unsaved_project)) {
+      const cachedLocaleMatches = cachedProject?.locale === effectiveLocale;
```
patch0: Can we have a little comment here that at the moment this will never match because the `cachedProject` doesn't have a `locale` attribute (yet), so this will always be false, which effectively disables the whole caching mechanism in the line below

Issue #1174, opened 2025-01-24
Missing styles on editor web component (projects site)


maxelkins:
## Overview

Recently I notice that some of styling on the editor web component was now broken on the projects site. I'm not sure if some of this happened when editor-ui were compeletely split.

Here is a quick visual of the errors - see in more detail in [Figjam](https://www.figma.com/board/O4TkEXljPGkS52278ioHd3/Projects-site-editor---style-review?node-id=0-1&t=dewNzNFR8Yh4K8QJ-1):

![Image](https://github.com/user-attachments/assets/eb06d4db-1ceb-49c7-a559-af2d3d19af53)

## Thoughts on the problem

I believe that the majority of these issue are because the CSS custom properties on the editor web component (used for theming) are currently unset in the projects site. However, being unset should not cause the styles to break - possibly this is because the custom properties have not been set with default fallbacks? Or they are not applying for some reason.

For example, `--sidebar-border` is set to `var(--rpf-grey-150)` in the editor [here](https://github.com/RaspberryPiFoundation/editor-ui/blob/1eae407c8253fd8618fcb2f4527ad64533a8472f/src/assets/stylesheets/index.scss#L44). In the projects site, this is unset.

## Solution

 - All themable custom properties should have a default fallback, so that if the consuming site not setting this custom properties, it will appear in the default theme.

maxelkins:
The general issue has still not been address but we currently default to editor styles for local dev with https://github.com/RaspberryPiFoundation/editor-ui/pull/1319

PR #1378, opened 2026-03-11
Update the default code editor styles

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/editor-ui/issues/1174

Light mode default -
image

Dark mode default -
image

Light mode editor styles -
image

Dark mode editor styles -
image


jamiebenstead:
> Does this make the `use_editor_styles` flag redundant? If it does, should we remove it as part of this?

It doesn't, but it should. The plan is to sort this in future work

PR #1407, opened 2026-03-27
Use cross-origin HTML preview renderer

mwtrew:
This moves preview rendering to an iframe with a document loaded cross-origin as a source. This enforces cross-origin separation and gives the preview window it's own localstorage, isolated from the parent. Content is injected into the doc via `postMessage()`. 

Most of the logic in the new `HtmlRenderer` component (besides message handling) has been moved over from `HtmlRunner`.


src/components/Editor/Runners/HtmlRunner/HtmlRunner.test.js
```diff
@@ -3,7 +3,7 @@ import { render, screen } from "@testing-library/react";
 import React from "react";
 import { Provider } from "react-redux";
 import HtmlRunner from "./HtmlRunner";
-import { codeRunHandled, triggerCodeRun } from "../../../../redux/EditorSlice";
+import { triggerCodeRun } from "../../../../redux/EditorSlice";
```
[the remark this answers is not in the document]
mwtrew: Not removed, moved to https://github.com/RaspberryPiFoundation/learner-experience-integration-tests/pull/34/changes.

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -229,189 +172,36 @@ function HtmlRunner() {
...
+      if (message?.type === MSG_HTML_PREVIEW_READY) {
+        setRendererReady(true);
       }
-      srcNode.setAttribute(attr, src);
-      srcNode.setAttribute("crossorigin", true);
-    });
-  };
+    },
+    [setRendererReady],
+  );
```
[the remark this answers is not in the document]
mwtrew: This could come from a number of sources, and is not sensitive.

src/components/Editor/Runners/HtmlRunner/HtmlRenderer.jsx
```diff
@@ -0,0 +1,200 @@
...
+      window.removeEventListener("message", handleEventFromPreview);
+    };
+  }, [handlePreviewUpdateFromHost]);
+
+  return (
+    <div className="htmlrenderer-root">
+      <iframe
+        className="htmlrunner-iframe"
+        title="preview-sandbox"
+        srcDoc={previewHtml ?? ""}
```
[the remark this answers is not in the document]
mwtrew: Works fine, so I don't think this is correct.

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
```diff
@@ -229,189 +172,36 @@ function HtmlRunner() {
...
+
+      output.current.contentWindow.postMessage(
+        {
+          type: MSG_HTML_PROJECT_UPDATE,
+          code: projectCode,
+          media: projectMedia,
+          current: indexPage.toString(),
+        },
+        process.env.HTML_RENDERER_URL,
+      );
```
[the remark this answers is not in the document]
mwtrew: Although this is correct (and I hadn't noticed it), this does work because the `targetOrigin` is parsed and the origin component extracted - see https://html.spec.whatwg.org/multipage/web-messaging.html#window-post-message-steps. I'm quite happy to leave it.

src/components/Editor/Runners/HtmlRunner/HtmlRenderer.jsx
```diff
@@ -0,0 +1,200 @@
...
+      );
+    } else if (matchingRegexes(allowedExternalLinks, srcNode.attrs[attr])) {
+      src = srcNode.attrs[attr];
+    }
+    srcNode.setAttribute(attr, src);
+    srcNode.setAttribute("crossorigin", true);
+  });
+};
+
+export function HtmlRenderer() {
```
zetter-rpf, line 129: I can't see any unit test coverage for this file, can we add tests for the most important behaviour? Some of this was previously tested in the HtmlRunner test file so maybe this could be moved across.

src/components/Editor/Runners/HtmlRunner/HtmlRunner.test.js
```diff
@@ -242,350 +239,6 @@ describe("When page does not exist", () => {
   });
 });
 
-describe("When run is triggered", () => {
```
zetter-rpf, line 245: These are the tests I mentioned in my other comment that could be moved to HtmlRender (except for the localstorage ones?)

src/components/Editor/Runners/HtmlRunner/HtmlRunner.jsx
zetter-rpf, line 204: Posting the `MSG_HTML_PROJECT_UPDATE` message is now important behaviour of this file, it would be useful to have a unit test that covered it.

zetter-rpf:
I was thinking about the seperate webpack config and tasks to run and build.

Is the intention to have the iframe deployed and served from the same bucket as editor-ui?
If it is, I wondered if there is value in having the seperate webpack config (instead of just a new entrypoint in the existing config). There is a benefit to having the iframe served from a different port in development, but it does add complexity to starting the server. Having the same config will still mean that we can test having the iframe on a separate port in development when included from standalone, it's only the local editor test page that won't be the same.

public/index-html-renderer.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
zetter-rpf: This might be here by mistake -there's another in src/

.env.example
```diff
@@ -3,8 +3,9 @@ REACT_APP_SENTRY_DSN=''
 REACT_APP_SENTRY_ENV='local'
 PUBLIC_URL='http://localhost:3011'
 ASSETS_URL='http://localhost:3011'
+HTML_RENDERER_URL='http://localhost:3003'
```
zetter-rpf: Example value should be :3011, and/or we could remove this in favour of using ASSETS_URL?

webpack.html-renderer.config.js
```diff
@@ -0,0 +1,98 @@
+const path = require("path");
```
zetter-rpf: Should be deleted?

zetter-rpf (approved):

PR #1420, opened 2026-04-07
1258 refactor cypress tests to use helpers

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1258

### DIFF
--- a/cypress/e2e/spec-wc-pyodide.cy.js
+++ b/cypress/e2e/spec-wc-pyodide.cy.js
@@ -5,7 +5,6 @@ import {
   getProgramInput,
   getPyodideOutput,
   getPythonConsoleOutput,
-  getSkulptRunner,
   getTextOutputTab,
   makeNewFile,
   runCode,
@@ -42,7 +41,7 @@ describe("Running the code with pyodide", () => {
   it("runs a simple program", () => {
     runCode('print("Hello world")');
     getPyodideOutput().should("not.contain.text", "Visual output");
-    getSkulptRunner().should("contain", "Text output");
+    getPyodideOutput().should("contain", "Text output");
 
     getPythonConsoleOutput().should("contain", "Hello world");
   });
--- a/cypress/e2e/spec-wc-pyodide.cy.js
+++ b/cypress/e2e/spec-wc-pyodide.cy.js
@@ -5,6 +5,7 @@ import {
   getProgramInput,
   getPyodideOutput,
   getPythonConsoleOutput,
+  getSkulptRunner,
   getTextOutputTab,
   makeNewFile,
   runCode,
@@ -41,9 +42,8 @@ describe("Running the code with pyodide", () => {
   it("runs a simple program", () => {
     runCode('print("Hello world")');
     getPyodideOutput().should("not.contain.text", "Visual output");
-    getPyodideOutput()
-      .find(".react-tabs__tab--selected")
-      .should("contain", "Text output");
+    getSkulptRunner().should("contain", "Text output");
+
     getPythonConsoleOutput().should("contain", "Hello world");
   });
 
--- a/cypress/e2e/spec-wc-pyodide.cy.js
+++ b/cypress/e2e/spec-wc-pyodide.cy.js
@@ -6,7 +6,6 @@ import {
   getPyodideOutput,
   getPythonConsoleOutput,
   getTextOutputTab,
-  getVisualOutputTab,
   makeNewFile,
   runCode,
   runProject,
@@ -41,7 +40,7 @@ describe("Running the code with pyodide", () => {
 
   it("runs a simple program", () => {
     runCode('print("Hello world")');
-    getVisualOutputTab().should("not.exist");
+    getPyodideOutput().should("not.contain.text", "Visual output");
     getPyodideOutput()
       .find(".react-tabs__tab--selected")
       .should("contain", "Text output");
--- a/cypress/e2e/spec-wc-pyodide.cy.js
+++ b/cypress/e2e/spec-wc-pyodide.cy.js
@@ -1,8 +1,17 @@
 import {
-  getErrorMessage,
   getEditorShadow,
+  getErrorMessage,
+  getFileButtonByName,
+  getProgramInput,
   getPyodideOutput,
   getPythonConsoleOutput,
+  getTextOutputTab,
+  getVisualOutputTab,
+  makeNewFile,
+  runCode,
+  runProject,
+  setCodeEditorContent,
+  stopProject,
 } from "../helpers/editor.js";
 
 const origin = "http://localhost:3011/web-component.html";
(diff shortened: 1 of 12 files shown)


cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
+    .should("not.be.null")
+    .should(($body) => {
+      expect($body.textContent || "").to.contain(text);
+    });
+
+export const expectHtmlRunnerPreviewToNotContainText = (text) =>
+  getHtmlRunnerBody().should("not.contain.text", text);
+
+export const expectErrorModalToNotExist = () =>
+  getEditorShadow().contains("An error has occurred").should("not.exist");
```
DNR500: I’d be a bit careful with this - it can become an anti-pattern in Cypress.

- contains("...") means “go find this element”
- should("not.exist") means “this element shouldn’t be there”

So they’re slightly at odds:

- contains(...) expects the element to exist
- should("not.exist") expects it not to exist

SO sometimes Cypress will fail on the contains(...) step before it even evaluates the assertion, which can make tests flaky or harder to reason about.

Simple rule to follow:

- Use contains(...) when you expect something to exist
- Use find(...) (or similar) + should("not.exist") when you expect something not to exist

Also worth checking other helpers in this file that return contains(...) — if they’re later used with should("not.exist"), they could run into the same issue.

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
+  getAddFileNameInput().clear().type(filename);
+  confirmAddFile();
+};
+
+export const runCode = (code) => {
+  // Add extra newline to ensure any tooltips are dismissed
+  setCodeEditorContent(`${code}\n`);
+  getRunButton().should("not.be.disabled");
+  runProject();
+};
```
DNR500, line 155: Really like the use of "flows" here. This is one of those things that can make your tests easy to work with and reason about.

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
 export const getErrorMessage = () =>
   getEditorShadow().find(".error-message__content");
 
-export const getSidebarPanel = () =>
-  getEditorShadow().findByTestId("sidebar__panel");
+export const getTextOutputTab = () =>
+  getPyodideOutput().findByLabelText("Text output");
+
+export const getVisualOutputTab = () =>
+  getPyodideOutput().contains(".react-tabs__tab", "Visual output");
```
DNR500: Might be good to see how this is used - I think it might get used with `.should("not.exist")`

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
+
+export const getSkulptSelectedTab = () =>
+  getSkulptRunner().find(".react-tabs__tab--selected");
+
+export const getSkulptTabByName = (name) => getSkulptRunner().contains(name);
```
DNR500: I think there’s an opportunity to tighten this up a bit. Since we’re already using react-tabs in places, we could probably lean on its accessible roles instead.

Might look something like..
```
getSkulptRunner().findByRole("tab", { name });
```

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
+  getEditorShadow().findByRole("button", { name: "Download project" });
+
+export const getAddFileButton = () =>
+  getEditorShadow().findByRole("button", { name: "Add file" });
+
+export const getConfirmAddFileButton = () =>
+  getEditorShadow().findAllByRole("button", { name: "Add file" }).last();
+
+export const getAddFileNameInput = () =>
+  getEditorShadow().findByRole("textbox");
```
DNR500: This feels a bit broad — findByRole("textbox") will match any textbox in the shadow root. If another textbox gets added there is a risk the test could end up typing into the wrong field.

Is there any defined `aria-label` or a label associate with the input? We just need a way to  specifically identify the input.

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
-export const getSidebarPanel = () =>
-  getEditorShadow().findByTestId("sidebar__panel");
+export const getTextOutputTab = () =>
+  getPyodideOutput().findByLabelText("Text output");
+
+export const getVisualOutputTab = () =>
+  getPyodideOutput().contains(".react-tabs__tab", "Visual output");
+
+export const getSkulptSelectedTab = () =>
+  getSkulptRunner().find(".react-tabs__tab--selected");
```
DNR500: might be able to try a role based way to select this, see comment below. Something like this might work...
```
findByRole("tab", { selected: true })
```

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
+export const getAddFileNameInput = () =>
+  getEditorShadow().findByRole("textbox");
+
+export const getFileButtonByName = (filename) =>
+  getEditorShadow().findByRole("button", { name: filename });
+
+export const getSettingsButton = () =>
+  getEditorShadow().find("[title='Settings']").first();
+
+export const getProgramInput = () => getEditorShadow().find("#input");
```
DNR500: Anything more specific for this one? if it has aria-label/label could try something like...
```
findByRole("textbox", { name: ... }) // if label exists
```

cypress/e2e/spec-wc-pyodide.cy.js
```diff
@@ -40,9 +41,7 @@ describe("Running the code with pyodide", () => {
 
   it("runs a simple program", () => {
     runCode('print("Hello world")');
-    getPyodideOutput()
-      .contains(".react-tabs__tab", "Visual output")
-      .should("not.exist");
+    getVisualOutputTab().should("not.exist");
     getPyodideOutput()
       .find(".react-tabs__tab--selected")
```
DNR500: You might be able to use findByRole and look for tab as part of this too

cypress/helpers/editor.js
```diff
@@ -1,18 +1,171 @@
...
+// Modal queries
+
+export const getErrorModalTitle = () =>
+  getEditorShadow().findByRole("heading", { name: "An error has occurred" });
+
+// Test output
+
+export const getResults = () => cy.get("#results");
+
+// Actions
```
DNR500, line 104: I really like how you have defined the groups like Layout, Actions, Assertions. This file will likely grow over time so maybe groups can be pulled out into their own files? What do you think?

DNR500 (commented):
This is looking a lot better! Its really good to see our use of CSS to select elements being remove. 

Few points on here - I think the main ones that we should address the use of `contains` and `should("not.exist")` and some of the selectors that are a little bit broad.

Give me a shout if you want to discuss anything on here

DNR500 (approved):
Nice! This makes reading though those tests a lot easier and good to be moving away from CSS selectors

PR #1421, opened 2026-04-08
Make using latest release easier

zetter-rpf:
Related to: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1245

This updates a 'latest_version' file in the bucket when editor ui is released. We'll use this file when editor-standalone is deployed to make sure we're on the latest version.

See commits for more.


adrian-rpf (approved):
LGMT

PR #1428, opened 2026-04-10
Set turtle graphics inside runCode to avoid null.firstChild

cocomarine:
Closes [Intermittent ExternalError when running Turtle projects after saving #1261](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1261)

### Findings
- The 'firstChild' error is not at all reproducible locally 
- In staging, sometimes it errors and sometimes not. Couldn't see any pattern 🤷‍♀️ 
- Upon inspecting `turtle.js` within the skupt module, it seemed that the type error could be coming from the `target` being null while checking `target.firstChild` for the while loop:
```js
// turtle.js 
...
function getConfiguredTarget() {
    var selector, target;

    selector = (Sk.TurtleGraphics && Sk.TurtleGraphics.target) || "turtle",
    target   = typeof selector === "string" ?
        document.getElementById(selector) :
        selector;
    // ensure that the canvas container is empty
    while (target.firstChild) {
        target.removeChild(target.firstChild);
    }
    return target;
}
...
```
- `SkulptRunner` runs runCode() in a parent useEffect. `VisualOutputPane` sets `Sk.TurtleGraphics.target` in a child useEffect. React runs parent effects before child effects, so from turtle import * can run before the child has set `Sk.TurtleGraphics.target`. 

### Potential fix
- Ensured `Sk.TurtleGraphics.target` points at `#turtleOutput` before turtle runs by setting it within the SkulpRunner's runCode()


DNR500 (approved):
Nice! I can see how this should avoid a race condition between turtle setup and code execution in VisualOutputPane. Also, thanks for adding the comments and tests

PR #1431, opened 2026-04-14
Issues/1263 scope assets to project in scratch

abcampo-iry:
Partially closes: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1263
This change makes Scratch keep sending the correct project id in its request metadata.
Scratch now updates the X-Project-ID metadata used by scratchFetch when the active Scratch project id changes.

Changes:

Added X-Project-ID to the allowed CORS/dev headers.
Now:
- stores the current Scratch project id
- sets X-Project-ID metadata when Scratch storage is initialised
- updates that metadata when Scratch changes to a new project id
- keeps sending Authorization metadata as before
Added tests related.

### DIFF
--- a/src/scratch.jsx
+++ b/src/scratch.jsx
@@ -35,12 +35,7 @@ const postScratchGuiEvent = (type, payload = {}) => {
   window.parent.postMessage({ type, ...payload }, allowedParentOrigin);
 };
 
-let scratchProjectId = projectId;
-let setScratchProjectMetadata = () => {};
-
 const handleUpdateProjectId = (updatedProjectId) => {
-  scratchProjectId = updatedProjectId;
-  setScratchProjectMetadata();
   postScratchGuiEvent("scratch-gui-project-id-updated", {
     projectId: updatedProjectId,
   });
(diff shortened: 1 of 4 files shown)


DNR500 (approved):

src/scratch.jsx
```diff
@@ -88,6 +93,15 @@ if (!projectId) {
...
       : "[scratch iframe] no scratch-gui-set-token message received before timeout";
 
+  setScratchProjectMetadata = () => {
+    if (!scratchFetchApi?.setMetadata) return;
+
+    const projectIdHeader =
+      scratchFetchApi.RequestMetadata?.ProjectId || "X-Project-ID";
+
+    scratchFetchApi.setMetadata(projectIdHeader, scratchProjectId);
+  };
```
zetter-rpf: I'm surprised we need this as I did test that this header gets set on asset uploading and fetching - Could you document what cases or for what API requests the project id header is not set? Is it to do with remixing?

abcampo-iry: Let me come back to this code, to double check, if this is required.

abcampo-iry: Not needed, so I will close this PR completely.

PR #1437, opened 2026-04-16
Update FilePanel to use design system button

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1297


DNR500 (approved):

zetter-rpf (changes requested):
Image

I saw that the deployed version of this has a different font for the file list - could that be from this change?

jamiebenstead:
> Image
> I saw that the deployed version of this has a different font for the file list - could that be from this change?

Good spot! Just pushed a commit to fix the font and the gap between icon and text

zetter-rpf (approved):
Nice one

PR #1438, opened 2026-04-16
1257: add CSP to scratch.html

DNR500:
issue: [1257](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1257)

I’ve built the CSP to try to tap into the env variables that we use for deployments.

I’ve also tried to do some due diligence on the testing staging and staging environments in order to anticipate any problems - I did find something that was a begin pulled in on those environments that I’ve added a workaround for [(see the comment below)](https://github.com/RaspberryPiFoundation/editor-ui/pull/1438#discussion_r3095172085)

### Local dev CSP output
This has to feature 'unsafe-eval' to allow the dev server to work
```
<meta http-equiv="Content-Security-Policy" content="
  default-src 'self';
  base-uri 'none';
  object-src 'none';
  script-src 'self' 'unsafe-inline' 'unsafe-eval';
  style-src 'self' 'unsafe-inline';
  worker-src 'self' blob:;
  child-src 'self' blob:;
  connect-src 'self' http://localhost:3009 http://localhost:3011;
  img-src 'self' data: blob: http://localhost:3011;
  media-src 'self' blob: http://localhost:3011;
  font-src 'self' data: http://localhost:3011;
  form-action 'self';
  upgrade-insecure-requests;
">
```

###  Staging CSP output (take values from deploy-main in ci-cd.yml) 
As the test environment uses editor-ui in staging that testing api needs to be listed on staging as well
```
<meta http-equiv="Content-Security-Policy" content="
  default-src 'self';
  base-uri 'none';
  object-src 'none';
  script-src 'self' 'unsafe-inline'  'unsafe-eval';
  style-src 'self' 'unsafe-inline';
  worker-src 'self' blob:;
  child-src 'self' blob:;
  connect-src 'self' https://staging-editor-api.raspberrypi.org https://test-editor-api.raspberrypi.org https://staging-editor-static.raspberrypi.org;
  img-src 'self' data: blob: https://staging-editor-static.raspberrypi.org;
  media-src 'self' blob: https://staging-editor-static.raspberrypi.org;
  font-src 'self' data: https://staging-editor-static.raspberrypi.org;
  form-action 'self';
  upgrade-insecure-requests;
">
```

###  Production (take values from deploy-tag in ci-cd.yml)
```
<meta http-equiv="Content-Security-Policy" content="
  default-src 'self';
  base-uri 'none';
  object-src 'none';
  script-src 'self' 'unsafe-inline'  'unsafe-eval';
  style-src 'self' 'unsafe-inline';
  worker-src 'self' blob:;
  child-src 'self' blob:;
  connect-src 'self' https://editor-api.raspberrypi.org https://editor-static.raspberrypi.org;
  img-src 'self' data: blob: https://editor-static.raspberrypi.org;
  media-src 'self' blob: https://editor-static.raspberrypi.org;
  font-src 'self' data: https://editor-static.raspberrypi.org;
  form-action 'self';
  upgrade-insecure-requests;
">
```
###  CSP violations for assets

Regarding current attempts to access images from `https://cdn.assets.scratch.mit.edu/` - the set up above will result in CSP errors in the console.    

CDN-ASSET-SCRATCH_CSP

This should be short lived until we arrive at the solution for asset hosting - I don’t think this will stop anything else in the app working as a results. If we choose to host images on a different domain to the ones already stated we should be able to just add that domain to the img-src/media-src CSP.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -168,7 +168,7 @@ jobs:
         if: env.AWS_SECRET_ACCESS_KEY != ''
         run: |
           aws s3 sync ./build/ s3://${{ secrets.AWS_S3_BUCKET }}/${{ needs.setup-environment.outputs.deploy_dir }} --endpoint ${{ secrets.AWS_ENDPOINT }} --progress-frequency 5
-          aws s3 sync ./build/chunks/ s3://${{ secrets.AWS_S3_BUCKET }}/chunks/ --endpoint ${{ secrets.AWS_ENDPOINT }} --exclude "*" --include "fetch-worker*" --include "mediapipe/*" --include "mediapipe/**" --progress-frequency 5
+          aws s3 sync ./build/chunks/ s3://${{ secrets.AWS_S3_BUCKET }}/chunks/ --endpoint ${{ secrets.AWS_ENDPOINT }} --exclude "*" --include "fetch-worker*" --include "mediapipe/**" --progress-frequency 5
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -66,6 +66,10 @@ on:
         required: false
         default: "https://staging-editor.raspberrypi.org,https://staging-editor-static.raspberrypi.org,https://test-editor.raspberrypi.org"
         type: string
+      csp_api_multiple_origins:
+        required: false
+        default: ""
+        type: string
 
     secrets:
       AWS_ACCESS_KEY_ID:
@@ -158,12 +162,13 @@ jobs:
           REACT_APP_SENTRY_DSN: ${{ inputs.react_app_sentry_dsn }}
           REACT_APP_SENTRY_ENV: ${{ inputs.react_app_sentry_env }}
           REACT_APP_ALLOWED_IFRAME_ORIGINS: ${{ inputs.react_app_allowed_iframe_origins }}
+          CSP_API_MULTIPLE_ORIGINS: ${{ inputs.csp_api_multiple_origins }}
 
       - name: Deploy site to S3 bucket
         if: env.AWS_SECRET_ACCESS_KEY != ''
         run: |
           aws s3 sync ./build/ s3://${{ secrets.AWS_S3_BUCKET }}/${{ needs.setup-environment.outputs.deploy_dir }} --endpoint ${{ secrets.AWS_ENDPOINT }} --progress-frequency 5
-          aws s3 sync ./build/chunks/ s3://${{ secrets.AWS_S3_BUCKET }}/chunks/ --endpoint ${{ secrets.AWS_ENDPOINT }} --exclude "*" --include "fetch-worker*" --progress-frequency 5
+          aws s3 sync ./build/chunks/ s3://${{ secrets.AWS_S3_BUCKET }}/chunks/ --endpoint ${{ secrets.AWS_ENDPOINT }} --exclude "*" --include "fetch-worker*" --include "mediapipe/*" --include "mediapipe/**" --progress-frequency 5
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -152,6 +152,7 @@ jobs:
       environment: staging
       react_app_base_url: ""
       react_app_sentry_env: staging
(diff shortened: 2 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -158,12 +162,13 @@ jobs:
...
           REACT_APP_SENTRY_ENV: ${{ inputs.react_app_sentry_env }}
           REACT_APP_ALLOWED_IFRAME_ORIGINS: ${{ inputs.react_app_allowed_iframe_origins }}
+          CSP_API_MULTIPLE_ORIGINS: ${{ inputs.csp_api_multiple_origins }}
 
       - name: Deploy site to S3 bucket
         if: env.AWS_SECRET_ACCESS_KEY != ''
         run: |
           aws s3 sync ./build/ s3://${{ secrets.AWS_S3_BUCKET }}/${{ needs.setup-environment.outputs.deploy_dir }} --endpoint ${{ secrets.AWS_ENDPOINT }} --progress-frequency 5
-          aws s3 sync ./build/chunks/ s3://${{ secrets.AWS_S3_BUCKET }}/chunks/ --endpoint ${{ secrets.AWS_ENDPOINT }} --exclude "*" --include "fetch-worker*" --progress-frequency 5
+          aws s3 sync ./build/chunks/ s3://${{ secrets.AWS_S3_BUCKET }}/chunks/ --endpoint ${{ secrets.AWS_ENDPOINT }} --exclude "*" --include "fetch-worker*" --include "mediapipe/**" --progress-frequency 5
```
DNR500, line 171: On testing and staging, but NOT on localhost, there is javascript being pulled in from the following CDN when trying to use the face sensing extension 

https://cdn.jsdelivr.net/npm/@mediapipe/face_detection@0.4.1646425229/face_detection_solution_simd_wasm_bin.js 

### Whats happening in scratch
In scratch-editor there is an initial attempt to see if chunks exist for that extension

[https://github.com/scratchfoundation/scratch-editor/blob/5283156f2c533d7c2783adc8622abed6d08002b9/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js#L94-L99](https://github.com/scratchfoundation/scratch-editor/blob/develop/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js#L94-L99) 

But a fallback is established if those chunks aren’t present locally

https://github.com/scratchfoundation/scratch-editor/blob/5283156f2c533d7c2783adc8622abed6d08002b9/packages/scratch-vm/src/extensions/scratch3_face_sensing/index.js#L101-L110

### What we can do in editor-ui to get around this
It looks like the scratch-editor makes these available as part of the build - so they should exist in the dist folder for scratch-gui In editor-ui, we are already copying the chunks from scratch-gui via web pack so this will include the face-sensing code

 https://github.com/RaspberryPiFoundation/editor-ui/blob/825e69b72b26df8acc4a69b8960eee562368730d/webpack.config.js#L18-L21  

https://github.com/RaspberryPiFoundation/editor-ui/blob/825e69b72b26df8acc4a69b8960eee562368730d/webpack.config.js#L246  

So updating and adding this to our aws sync should get this working on all our environments - essentially we would then be hosting the code ourselves and the can request wouldn’t happen.  

I’ve had to dig a little to understand the above - but I think this one liner should enable the face sensing extension without having to hit the CDN

zetter-rpf: Great. Since we already have disabled some extensions that is also an option if this there are more problems with this extension or others.

zetter-rpf:
I saw Scratch doesn't work for the build for this PR - https://staging-editor-static.raspberrypi.org/branches/1438_merge/web-component.html


> react-dom.production.min.js:2 Error: Error in Scratch GUI [location=https://staging-editor-static.raspberrypi.org/branches/1438_merge/scratch.html?project_id=cool-scratch.json&api_url=https%3A%2F%2Fstaging-editor-static.raspberrypi.org%2Fbranches%2F1438_merge&scratchMetadata=1&parent_origin=https%3A%2F%2Fstaging-editor-static.raspberrypi.org]: EvalError: Evaluating a string as JavaScript violates the following Content Security Policy directive because 'unsafe-eval' is not an allowed source of script: script-src 'self' 'unsafe-inline'".

DNR500:
> I saw Scratch doesn't work for the build for this PR - https://staging-editor-static.raspberrypi.org/branches/1438_merge/web-component.html
> 
> > react-dom.production.min.js:2 Error: Error in Scratch GUI [location=https://staging-editor-static.raspberrypi.org/branches/1438_merge/scratch.html?project_id=cool-scratch.json&api_url=https%3A%2F%2Fstaging-editor-static.raspberrypi.org%2Fbranches%2F1438_merge&scratchMetadata=1&parent_origin=https%3A%2F%2Fstaging-editor-static.raspberrypi.org]: EvalError: Evaluating a string as JavaScript violates the following Content Security Policy directive because 'unsafe-eval' is not an allowed source of script: script-src 'self' 'unsafe-inline'".

I've had look at this - in our current Scratch GUI bundle/runtime, `new Function(...) ` is executed, so `unsafe-eval` is currently required by the Scratch iframe CSP.

Adding `unsafe-eval` for all environment for now - I don't think there is an easy work around for this at the moment

zetter-rpf (approved):
> Adding unsafe-eval for all environment for now - I don't think there is an easy work around for this at the moment

Even with unsafe-eval this is really valuable - it will help us see in testing if we are depending on content from other domains.

PR #1439, opened 2026-04-17
Refactor to split scratch.jsx into clearer parts

zetter-rpf:
Related to: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1197

Refactor to split scratch.jsx now that it's responsibilities have grown and simplify tests

In https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1197 I'll need to update the access token that's passed into Scratch. This should make that easier.

I've kept `scratch.jsx` as responsible for initializing the page and setting up authentication. ScratchEditor is responsible for rendering the Scratch editor component and setting up the handlers it needs.

I've also extracted `WrappedScratchGui` as it makes the tests easier to set up when you don't need to mock the process of setting up the higher order component.

### DIFF
--- a/src/components/ScratchEditor/ScratchEditor.jsx
+++ b/src/components/ScratchEditor/ScratchEditor.jsx
@@ -1,0 +1,74 @@
+import scratchProjectSave from "../../utils/scratchProjectSave.js";
+import { useCallback } from "react";
+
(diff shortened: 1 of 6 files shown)


DNR500 (approved):
Thanks for this, this looks cleaner

PR #1442, opened 2026-04-20
1261: fix for turtle target reference error

DNR500:
issue: [1261](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1261)

### Context

`editor-ui` is embedded inside `editor-standalone` as a web component.

We hit a weird turtle failure when:
- first loading a project - works fine  
- navigating away (e.g. to a lesson page)  
- then coming back - breaks  

A full page refresh fixes it, so this is clearly something about remounting rather than initial load.

The error we were seeing:

```
Cannot read properties of null (reading 'firstChild')
```

At first I suspected this was something to do with lifecycle + Shadow DOM + how Skulpt tries to resolve the turtle target in the light DOM.

---

### What Skulpt is doing

When Python imports `turtle`, Skulpt runs `getConfiguredTarget()`:

- https://github.com/skulpt/skulpt/blob/master/src/lib/turtle.js#L4

It determines the target using:

- `Sk.TurtleGraphics.target`, or falls back to `"turtle"`  
- https://github.com/skulpt/skulpt/blob/master/src/lib/turtle.js#L7  

If it ends up with a string, it does:

```js
document.getElementById(selector)
```

- https://github.com/skulpt/skulpt/blob/master/src/lib/turtle.js#L9  

It then assumes that worked and immediately does:

```js
while (target.firstChild) { ... }
```

- https://github.com/skulpt/skulpt/blob/master/src/lib/turtle.js#L12  

---

### Why this breaks for us

Our turtle output lives inside a **Shadow DOM**, not the top-level document.

So if Skulpt falls back to `"turtle"`:
- `document.getElementById("turtle")` - returns `null`
- our actual element is `#turtleOutput`, and it’s not even in the light DOM anyway

So we end up with:

```js
target === null
```

and then:

```js
target.firstChild
```

So exactly the error we saw.

This lines up with what I saw while debugging:
- failures consistently happen on remount (after navigating away and back)
- if I redirect `"turtle"` lookup to the actual Shadow DOM element, everything works again

---

### Additional findings

I also tried tightening up our lifecycle handling:

- made sure turtle teardown/reinit happens before rerunning code  
- moved more of this into `SkulptRunner`  

After that:
- I can confirm `Sk.TurtleGraphics.target` **is correctly set** before `Sk.importMainWithBody(...)`

**But** Skulpt still seems to hit its internal fallback path and try `"turtle"` anyway.

So this feels like:
- either the target is getting lost/reset internally  
- or Skulpt is re-resolving it in a way that ignores what we set  

---

### The fix (for now)

Not pretty, and this doens't feel like a nice fix, but its the only thing I’ve found that reliably works:

- during execution, temporarily shim `document.getElementById`
- if it’s called with `"turtle"`, return our actual `#turtleOutput` element
- restore the original function in a `finally()` block

---

### Things we might consider

A couple of directions that *might* be worth exploring (more thinking out loud than firm suggestions):

Some possible changes to Skulpt.. 
- Skulpt probably shouldn’t assume `getElementById` will always succeed  
- It would be good if it consistently respected an explicitly provided `Sk.TurtleGraphics.target`, rather than falling back again  

Another option could be introducing something like a `targetFunc()`:

- instead of passing a static value, allow a function that resolves the element at runtime  
- this would play much more nicely with things like Shadow DOM and remounting, where the element isn’t always stable or in the light DOM  

Something along the lines of:

```js
Sk.TurtleGraphics.targetFunc = () => {
  return resolveTurtleElementSomehow();
};
```

That way Skulpt doesn’t need to guess or fall back — it just asks for the element when it needs it.

Not saying we should go and patch Skulpt right now, but it does feel like the kind of thing that could be improved upstream if we ever wanted to explore it - and it might be the only thing that gives a nice solution

### DIFF
--- a/src/components/Editor/Runners/PythonRunner/SkulptRunner/SkulptRunner.jsx
+++ b/src/components/Editor/Runners/PythonRunner/SkulptRunner/SkulptRunner.jsx
@@ -87,7 +87,6 @@ const SkulptRunner = ({ active, outputPanels = ["text", "visual"] }) => {
   const senseHatAlwaysEnabled = useSelector(
     (state) => state.editor.senseHatAlwaysEnabled,
   );
-  const projectImages = useSelector((state) => state.editor.project.image_list);
   const reactAppApiEndpoint = useSelector((s) => s.editor.reactAppApiEndpoint);
   const output = useRef();
   const dispatch = useDispatch();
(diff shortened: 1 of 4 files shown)


zetter-rpf (approved):
Thanks for the detailed write-up!

PR #1447, opened 2026-04-24
add loading loader to wc loader

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1295

### Notes
- Another loader from editor-standalone appears in the middle of the page before editor-ui loader appears
- Added a Loader component that is similar to the one in standalone and used it to show loading in the same place with a similar style as standalone
- Also added a LoadFailed component
- When 'Try again' is clicked, it reloads the page.
- Briefly looked into coordinating editor-ui and standalone loaders but decided to take a simpler route for now. 

### Screenshots
Screenshot 2026-04-24 at 14 43 07
Screenshot 2026-04-27 at 08 14 20

- When in dark theme:
Screenshot 2026-04-27 at 13 38 58
Screenshot 2026-04-27 at 14 00 30


maxelkins:
I think this is an improvement on the current failed state 👍 

### Potential iteration (not required if you just wanna ship it)

What normally fixes this? Is it just a refresh? Could we have:

"Load failed. [Try again](url)"

- Try again triggers a refresh
- Load failed can be the normal text colour

maxelkins:
@cocomarine Would make "Try again" the link colour (like on the [homepage](https://editor.raspberrypi.org/en/)) and then lets get this in. We can take a look in detail when we have more time 👍 

Thank you!

DNR500:
I think it might need an ever so slight spacing adjustment to match the one in standalone

editor-standalone
loader-standalone

editor-ui
loader-spacing-editor-ui

Might also need change the text color when dark mode too? 
loader-editor-ui-dark

Would be nice at some point to move this into the design system so we can more easily keep this consistent

DNR500 (approved):
Dropped a comment on with some suggestions for smaller changes - but this is going to look a lot better. Thanks for this

PR #1456, opened 2026-05-07
Fix the HTML preview locally

zetter-rpf:
Found while working on https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1379

In production we have a cloudflare rule that allows cross-origin requests. Without this the iframe does not render in locally (tested in Chrome).


zetter-rpf:
@abcampo-iry thought you would be a good reviewer as you had the same problem

mwtrew (approved):

PR #1460, opened 2026-05-08
1344: remove unneeded plausible env vars

DNR500:
Relates to [1344](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1344)

This PR does **NOT** change analytics behaviour in `editor-ui` -  it removes environment variables that are not wired into `editor-ui `analytics and could cause confusion.

### What changed
- No functional changes were introduced to Plausible event names or event trigger points in `editor-ui`.
- We validated that key UI actions are still calling window.plausible(...) as expected.

Why the env vars aren’t needed
- Analytics bootstrap/config changes are handled in `editor-standalone`.
- `editor-ui` event emission continues to work through that integration, using window.plausible() provided by `editor-standalone`.
- The env vars removed in this PR are not actually connected to `editor-ui` event dispatch. 

What this means
- `editor-ui` remains the source of UI event calls.
- `editor-standalone` remains responsible for loading/configuring Plausible.
- This preserves existing tracking behaviour while reducing config ambiguity.

Checks
* Confirmed `window.plausible(...)` calls fire from editor-ui actions in local development, using the window.plausible function established by `editor-standalone`.


zetter-rpf (approved):
Thanks for tidying up!

PR #1464, opened 2026-05-13
Cache fixing by unit path invalidation instead of individual file listing

abcampo-iry:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1371

#### What changed

  This PR changes the deploy workflow to purge Cloudflare by deployed path prefix instead of by a hard-coded list of individual files.

A staging deploy now purges:
```bash
  {
    "prefixes": [
      "staging-editor-static.raspberrypi.org/branches/main/"
    ]
  }
```

while before:

```bash
  {
    "files": [
      "https://staging-editor-static.raspberrypi.org/branches/main/web-component.html",
      "https://staging-editor-static.raspberrypi.org/branches/main/web-component.js",
      "https://staging-editor-static.raspberrypi.org/branches/main/scratch.html",
      "https://staging-editor-static.raspberrypi.org/branches/main/scratch.js",
      "https://staging-editor-static.raspberrypi.org/latest_version",
      "https://staging-editor-static.raspberrypi.org/branches/main/html-renderer.html",
      "https://staging-editor-static.raspberrypi.org/branches/main/html-renderer.js"
    ]
  }
```

The PR also uploads latest_version with Cache-Control: no-cache, because that file is intentionally updated in place.
For production tags, before it purged individual files inside one release plus latest_version. Now it purges:
```bash


  {
    "prefixes": [
      "editor-static.raspberrypi.org/releases/<tag>/",
      "editor-static.raspberrypi.org/latest_version"
    ]
  }
```

## Why

Staging should serve the latest deployed version of editor-ui after each deploy.

The previous single-file purge targeted paths such as `web-component.js`, `scratch.js`, and `html-renderer.js`. That can be unreliable because Cloudflare may cache separate variants of the same URL when request headers differ, for example with different Origin headers. If the purge request does not match the cached variant, Cloudflare can continue serving an older file. Prefix purging clears the deployed build path as a unit. The next request after deploy fetches the updated files from the bucket, and Cloudflare then caches those files again as normal.
This is a targeted deploy/cache invalidation fix.

References:
- Purge cache by prefix (https://developers.cloudflare.com/cache/how-to/purge-cache/purge_by_prefix/)
  - Cloudflare documents prefix purge for clearing a directory/path, reducing purge calls, and causing the next request to refetch from origin.
- Purge Cached Content API (https://developers.cloudflare.com/api/resources/cache/methods/purge/)
- Cache Keys (https://developers.cloudflare.com/cache/how-to/cache-keys/)
- Purge by single-file (https://developers.cloudflare.com/cache/how-to/purge-cache/purge-by-single-file/)
- Origin Cache Control (https://developers.cloudflare.com/cache/concepts/cache-control/)
- Browser Cache TTL (https://developers.cloudflare.com/cache/how-to/edge-browser-cache-ttl/set-browser-ttl/)

Idea: 
Longer term, we should move to content-hashed asset filenames with immutable caching, leaving only small stable entrypoints or manifests as revalidated files.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -172,7 +172,8 @@ jobs:
             exit 0
           fi
 
-          echo -n "${{ needs.setup-environment.outputs.deploy_dir }}" | aws s3 cp - s3://${{ secrets.AWS_S3_BUCKET }}/latest_version --endpoint ${{ secrets.AWS_ENDPOINT }} --content-type "text/plain"
+          # Use no-cache so latest_version is revalidated after tag deploys.
+          echo -n "${{ needs.setup-environment.outputs.deploy_dir }}" | aws s3 cp - s3://${{ secrets.AWS_S3_BUCKET }}/latest_version --endpoint ${{ secrets.AWS_ENDPOINT }} --content-type "text/plain" --cache-control "no-cache"
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -181,11 +181,32 @@ jobs:
...
+
+          echo "Purging Cloudflare cache prefixes:"
+          jq -r '.prefixes[] | " - \(.)"' <<<"$purge_payload"
+
           response="$(
             curl -sS --fail-with-body -X POST "https://api.cloudflare.com/client/v4/zones/${CLOUDFLARE_ZONE_ID}/purge_cache" \
               -H "Authorization: Bearer ${CLOUDFLARE_API_TOKEN}" \
               -H "Content-Type: application/json" \
-              --data '{"files":["${{ needs.setup-environment.outputs.public_url }}/web-component.html", "${{ needs.setup-environment.outputs.public_url }}/web-component.js", "${{ needs.setup-environment.outputs.public_url }}/scratch.html", "${{ needs.setup-environment.outputs.public_url }}/scratch.js", "${{ inputs.base_url }}/latest_version", "${{ needs.setup-environment.outputs.html_renderer_url}}/html-renderer.html", "${{ needs.setup-environment.outputs.html_renderer_url}}/html-renderer.js"]}'
+              --data "$purge_payload"
```
zetter-rpf: I do find this code harder to understand now because it is doing a lot now and the has all the actions interpolation within it. Do you think extracting it to a bash or Ruby script could make it easier to follow? Or is there any other way to simplify it?

zetter-rpf:
Thanks for the explanation and links to the doc. I've added a comment. 

I also wondered if we set a no cache for `latest-version`, do we need to purge it any more?

abcampo-iry:
> Thanks for the explanation and links to the doc. I've added a comment.
> 
> I also wondered if we set a no cache for `latest-version`, do we need to purge it any more?

he purge is what immediately clears Cloudflare after deploy. The no-cache header is still useful because latest_version is mutable, so any cache that stores it should revalidate before reuse. It is not strictly required if purge always works, but it is the safer cache policy for this file.

.github/workflows/deploy.yml
abcampo-iry: Tried to simplify

zetter-rpf (approved):
Great, and thanks for improving the readability, I found that a lot easier to follow

PR #1465, opened 2026-05-13
Use cooldowns to reduce the risk of supply chain attacks

zetter-rpf:
Related to a similar change in editor-standalone: https://github.com/RaspberryPiFoundation/editor-standalone/pull/881

This does three things:

- Set up dependabot so we get non-security updates to, will a cooldown
- Add a cooldown to yarn using the `npmMinimalAgeGate` setting.
- Make docker's use of yarn more consistent with other environments

See commits for more

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,5 +2,10 @@ build/
 dist/
 .git/
 node_modules/
-.yarn/
-.yarnrc.yml
+
+.yarn/*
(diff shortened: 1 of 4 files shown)


abcampo-iry (approved):
This looks reasonable

DNR500 (approved):

PR #1469, opened 2026-05-14
Mimic autosave in Scratch Editor

abcampo-iry:
Closes: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1212

This change adds autosave for Scratch projects and keeps the behaviour close to the existing Python and HTML editors.

Python and HTML already autosave after a short delay when the user owns a saved project. Scratch is different because project changes happen inside the Scratch iframe, so we listen to Scratch VM change events and pass those changes back to the main app.

Autosave only starts once the Scratch project has been saved or remixed and has an identifier. New Scratch projects, and projects that need to be remixed first, still show the manual Save button. After that first save/remix, the manual Save button is hidden and the header shows the autosave status instead.

The autosave delay is 2 seconds. If a save is already running, we do not start another one immediately; we queue the next autosave until the current save finishes.

Scratch save events now update the shared Redux editor save state: saving, lastSavedTime, and lastSaveAutosave. This is the same state used by the existing save status UI, so Scratch can reuse SaveStatus instead of keeping a separate local save-status flow in the Scratch hook.

The Scratch save/remix tests have also been updated to cover the new flow: first save/remix, identifier update, hiding the manual Save button, and triggering autosave once the project is eligible.

As Teacher:

https://github.com/user-attachments/assets/ee9010bc-0ae8-4cb7-953e-eeaea70253ec

As Student (first time):

https://github.com/user-attachments/assets/60bb464b-6bb6-4fc9-9131-d9c84994496d

### DIFF
--- a/src/components/ScratchEditor/ScratchIntegrationHOC.jsx
+++ b/src/components/ScratchEditor/ScratchIntegrationHOC.jsx
@@ -8,6 +8,10 @@ import {
   setStageSize,
 } from "@scratch/scratch-gui";
 import { allowedIframeHost } from "../../utils/iframeUtils";
+import { postScratchGuiEvent } from "./events.js";
+
+const targetIdsFor = (targets = []) =>
+  targets.map((target) => target.id).join("|");
 
 const ScratchIntegrationHOC = function (WrappedComponent) {
   class ScratchIntegrationComponent extends React.Component {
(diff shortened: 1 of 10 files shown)


adrian-rpf:
I see no issues but due to the impact of this change should have an another engineer review. 

non-blocking: We should discuss how we are handling the auto-save time. [See Thread .](https://raspberrypifoundation.slack.com/archives/C02JBAA2NFP/p1778854571171979)

src/components/ScratchEditor/ScratchIntegrationHOC.jsx
```diff
@@ -19,13 +23,26 @@ const ScratchIntegrationHOC = function (WrappedComponent) {
       this.handleUpload = this.handleUpload.bind(this);
       this.handleRemix = this.handleRemix.bind(this);
       this.handleSave = this.handleSave.bind(this);
+      this.handleProjectChanged = this.handleProjectChanged.bind(this);
+      this.handleTargetsUpdate = this.handleTargetsUpdate.bind(this);
     }
     componentDidMount() {
       window.addEventListener("message", this.handleMessage);
+      this.props.vm?.on?.("PROJECT_CHANGED", this.handleProjectChanged);
+      this.props.vm?.on?.("targetsUpdate", this.handleTargetsUpdate);
```
zetter-rpf: What is a target in this context and when does it fire (and a project change doesn't?)

Also I can see your guarding this for when the `vm` doesn't exist, or it has no `on` function - why might this happen? May autosave not work in some cases?

zetter-rpf (approved):
Nice one,

I did find the changes in the project bar and hooks complex and haven't gone through them line by line, but I can see it works well and has good test coverage. 

I just had couple of questions to help my understanding that might be making sure are clearer in the commit/pr/code.

src/components/ScratchEditor/ScratchIntegrationHOC.jsx
abcampo-iry: I think you are right, on the "target in this context", I was thinking it needed a wider scope that's why i binded both I think I can make it with only project_changed.

On the on, its more defensive, but I would bet is probably ok without.

zetter-rpf: > On the on, its more defensive, but I would bet is probably ok without.

Sometimes being defensive is less helpful - it can mask unexpected errors and hide important states to cover. For example here, if `vm` didn't exist, the code would mean that this bit ran fine, but I wasn't sure if the auto save would work or if the user could save at all. If this could happen, it would be useful to have more tests around.

PR #1483, opened 2026-05-28
Add improved error messaging to python projects

cocomarine:
Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1448

### Changes
- Added pfem package and required changes for it to run properly
- Added dompurify package for sanitising HTML contents that are returned from pfem
- Added `friendly_errors_enabled` attribute set to `false` by default and added to redux
- The new redux state read by PythonRunner and passed as a prop to both PyodideRunner and SkulptRunner
- Added pfem logic to both runners and stored the friendly error html in redux
- Updated ErrorMessage component to display the friendly errors conditionally
  - For now, CSS used to restrict to displaying title and summary parts of friendly error html 
  - pfem will be updated so that we can choose parts before storing in redux
- ⚠️ The error messages will be formatted in another PR: https://github.com/RaspberryPiFoundation/editor-ui/pull/1485

### Behaviour

- When the editor-wc in the host app has `friendly_errors_enabled` attribute not set or set to `false`, it shows the original error.
   Screenshot 2026-05-28 at 14 59 03

- When the editor-wc in the host has `friendly_errors_enabled` attribute set to `true`, it shows the original error followed by title and summary of the new friendly error.
  Screenshot 2026-06-01 at 15 16 33

### Notes on using the pfem ESM in Jest
- Tried adding it to `jest.transformIgnorePatterns` in package.json, but resulted in a cascade of errors from plotly
  - Just moved `three` to  `jest.transformIgnorePatterns` from `scripts.test` 
- Added global jest mocks for pfem and plotly to `setupTests.js`
  - Also imported plotly to `PyodideRunner/VisualOutputPane.test.js` 
- <- with a CJS build of pfem, these changes may be unnecessary

### DIFF
--- a/src/components/Editor/ErrorMessage/ErrorMessage.jsx
+++ b/src/components/Editor/ErrorMessage/ErrorMessage.jsx
@@ -20,16 +20,14 @@ const ErrorMessage = () => {
       {friendlyError && (
         <div className="error-message__friendly">
           {friendlyError.title && (
-            <p
-              className="error-message__friendly-title"
-              dangerouslySetInnerHTML={{ __html: friendlyError.title }}
-            />
+            <p className="error-message__friendly-title">
+              {friendlyError.title}
+            </p>
           )}
           {friendlyError.summary && (
-            <p
-              className="error-message__friendly-summary"
-              dangerouslySetInnerHTML={{ __html: friendlyError.summary }}
-            />
+            <p className="error-message__friendly-summary">
+              {friendlyError.summary}
+            </p>
           )}
         </div>
       )}
--- a/src/components/Editor/ErrorMessage/ErrorMessage.jsx
+++ b/src/components/Editor/ErrorMessage/ErrorMessage.jsx
@@ -1,6 +1,7 @@
 import React, { useContext, useEffect, useRef } from "react";
 import "../../../assets/stylesheets/ErrorMessage.scss";
 import { useSelector } from "react-redux";
+import DOMPurify from "dompurify";
 import { SettingsContext } from "../../../utils/settings";
 
 const ErrorMessage = () => {
@@ -14,22 +15,19 @@ const ErrorMessage = () => {
       message.current.innerHTML = error;
     }
   }, [error]);
+
+  const friendlyErrorHtml = friendlyError?.html
+    ? DOMPurify.sanitize(friendlyError.html)
+    : null;
+
   return error ? (
     <div className={`error-message error-message--${settings.fontSize}`}>
-      <pre ref={message} className="error-message__content"></pre>
-      {friendlyError && (
-        <div className="error-message__friendly">
-          {friendlyError.title && (
-            <p className="error-message__friendly-title">
-              {friendlyError.title}
-            </p>
-          )}
-          {friendlyError.summary && (
-            <p className="error-message__friendly-summary">
-              {friendlyError.summary}
-            </p>
-          )}
-        </div>
+      <pre ref={message} className="error-message__content" />
+      {friendlyErrorHtml && (
+        <div
+          className="error-message__friendly"
+          dangerouslySetInnerHTML={{ __html: friendlyErrorHtml }}
+        />
       )}
     </div>
   ) : null;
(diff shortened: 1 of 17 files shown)


maxelkins:
~~@cocomarine Just so you know, we are still likely to want to show the normal error message too as well as the friendly one.~~

webpack.config.js
```diff
@@ -67,6 +67,12 @@ const moduleRules = [
     exclude: /node_modules/,
     use: ["babel-loader"],
   },
+  {
+    test: /\.js$/,
+    include:
+      /node_modules\/@raspberrypifoundation\/python-friendly-error-messages/,
+    resolve: { fullySpecified: false },
```
mwtrew, line 74: Why is this necessary?

src/components/Editor/Runners/PythonRunner/PyodideRunner/VisualOutputPane.test.js
```diff
@@ -140,7 +142,11 @@ describe("when there is plotly output", () => {
   });
 
   test("it renders the plotly chart as an svg", () => {
-    expect(screen.getByText("Test Plot")).toBeInTheDocument();
+    expect(Plotly.newPlot).toHaveBeenCalledWith(
```
mwtrew, line 144: This isn't quite as strong an expectation as before - just because it's been called doesn't mean the chart is on-screen.

src/components/Editor/ErrorMessage/ErrorMessage.jsx
```diff
@@ -1,21 +1,34 @@
...
 
 const ErrorMessage = () => {
   const message = useRef();
   const error = useSelector((state) => state.editor.error);
+  const friendlyError = useSelector((state) => state.editor.friendlyError);
   const settings = useContext(SettingsContext);
 
   useEffect(() => {
     if (message.current) {
       message.current.innerHTML = error;
```
mwtrew: I'm not sure why `message` was a ref, but is seems a bit odd for the "unfriendly" path to use a ref and not the friendly path. Maybe we can do what you've done with `friendlyError` for both error types and remove this?

src/components/Editor/Runners/PythonRunner/PyodideRunner/PyodideRunner.jsx
```diff
@@ -33,9 +40,22 @@ const getWorkerURL = (url) => {
...
+  friendlyErrorsEnabled = false,
+}) => {
   const [pyodideWorker, setPyodideWorker] = useState(null);
 
+  useEffect(() => {
+    if (friendlyErrorsEnabled) {
+      loadCopydeckFor(navigator.language, {
+        base: `${process.env.PUBLIC_URL}/python-error-copydecks/`,
+      });
+      registerAdapter("pyodide", pyodideAdapter);
```
mwtrew, line 55: I haven't looked in to what `registerAdapter` does, but it sounds like the sort of thing that _might_ have a corresponding teardown that the `useEffect` could run as a [cleanup function](https://react.dev/reference/react/useEffect#parameters).

mwtrew (approved):
Looking great in general! A few comments/questions but I'm happy for it to go in.

webpack.config.js
cocomarine: Without this, we get an error because the pfem package uses `./utils` instead of `./utils.js`. So I needed to disable ESM strict resolution just for pfem here.

src/components/Editor/Runners/PythonRunner/PyodideRunner/PyodideRunner.jsx
cocomarine: Registering pyodide runtime in the pfem, not in the component, so probably not needed here.

src/components/Editor/ErrorMessage/ErrorMessage.jsx
cocomarine: Thanks! Didn't look at `message` properly 😅 Now looking at it, you are right. Will update it.

PR #1485, opened 2026-05-29
Error message UI

maxelkins:
## Summary

- ~~Moves ErrorMessage to input side of editor (TBC)~~
- Adds FriendlyErrorMessage component to house the output from pfem
- Adds FriendlyErrorMessage to ErrorMessage (currently hidden)

## Notes

- Moving the error message here might be too much and confuse feedback - so might move out of this PR
- HJ: [new branch ](https://github.com/RaspberryPiFoundation/editor-ui/pull/1487) created by rebasing this to https://github.com/RaspberryPiFoundation/editor-ui/pull/1483

PR #1487, opened 2026-06-03
Improve friendly error message

cocomarine:
❗ This branch was created by rebasing https://github.com/RaspberryPiFoundation/editor-ui/pull/1485 to https://github.com/RaspberryPiFoundation/editor-ui/pull/1483. 

### How to enable friendly errors
- editor-standalone (for editor): add `friendly_errors_enabled: true` to `options` in src/components/Editor/Project/Project.jsx

### Decisions made upon adding [pfem v. 0.7.0](https://github.com/RaspberryPiFoundation/python-friendly-error-messages/releases)
- For now, we are showing `title` and `summary` of friendly errors (potentially `why` bit might also be added later on)
- `ErrorMessage` now optionally includes `FriendlyErrorMessage`
- The new attribute added and set to `true` in in src/web-component.html (for http://localhost:3011/web-component.html)
- Moved `ErrorMessage` to the bottom of input panel above Run button
   -  In mobile screens (`MobileProject`), when the error is triggered after run, the input panel becomes the selected tab so that users can see the error (if no error triggered, the output panel becomes selected)
   - For Skulpt runner, [the logic of showing error when not in output only mode is now removed](https://github.com/RaspberryPiFoundation/editor-ui/pull/1487/changes#diff-b905586781d8ed679facb77f80b18decb431bb0c1c9576b34e43745c3348a6ab), keeping the same behaviour of no error in output only mode. 
   - For Pyodide runner, [a logic for showing error message when in output only mode](https://github.com/RaspberryPiFoundation/editor-ui/pull/1487/changes#diff-9950a7b6007ef96b6e4ca4e8e0ad4461249437b7121f74511804b3b0139ada2a) is now added
      - As in `WebcomponentProject.jsx`, EditorInput is not rendered at all in output only mode, resulting in `ErrorMessage` displayed only by Output panel of Pyodide
- When there are multiple Python files:
   - An error from `main.py` shows first.
   - When there are errors only in files other than `main.py`, they will show (one at a time). 
   - Whatever error is shown will stay even when you select other tabs within the input panel. 
 
### WIP and future pfem 
- More discussion needed in terms of the possibility of adding `why` of friendly error, copy and UI

### Screen recordings of editor with friendly messages
- Multiple python files
  https://github.com/user-attachments/assets/a70727a4-121a-4c05-859f-f60f2cc29046

- Responsiveness
  https://github.com/user-attachments/assets/6697cfd4-fca9-4c04-af66-92b181a8e34d

- Mobile screen (< 600 px)
  https://github.com/user-attachments/assets/48736879-ff09-439d-8b1f-fb633a215333


src/components/Editor/Runners/PythonRunner/SkulptRunner/SkulptRunner.jsx
```diff
@@ -176,10 +174,15 @@ const SkulptRunner = ({
...
   useEffect(() => {
     if (friendlyErrorsEnabled) {
-      loadCopydeckFor(navigator.language, {
-        base: `${process.env.PUBLIC_URL}/python-error-copydecks/`,
-      });
-      registerAdapter("skulpt", skulptAdapter);
+      try {
+        loadCopydeckFor(navigator.language || "en", {
+          base: `${process.env.PUBLIC_URL}/python-error-copydecks/`,
+        });
```
zetter-rpf: We normally use locale from the i18n library rather than the navigator language - is this intentional here? I'm not sure if it will cause any problems - does `loadCopydeckFor` always fall back to en anyway?

src/web-component.html
```diff
@@ -106,17 +104,18 @@
...
-          .href
-          .replace(/\/$/, "");
+        const scratchApiEndpoint = new URL(
+          ".",
+          window.location.href,
+        ).href.replace(/\/$/, "");
         newWebComp.setAttribute("scratch_api_endpoint", scratchApiEndpoint);
-        
+
         queryParams.forEach((value, key) => {
```
zetter-rpf, line 120: I don't understand the reasons for these changes and a few others - do you have your linter set up correctly in your editor? Or is there a reason this file wasn't properly linted before?

Just makes it a bit harder to see what you actually have changed

zetter-rpf (approved):
Great, I like the tests you've added - clear and easy to see what they are testing.

The PR was harder to review because there are a lot of changes across different files  so it's harder for me to see that we're doing everything you said in the PR description - I don't think it's worth splitting out now, but for next time it could be helpful to think about how to break something like this up into logical commits and/or PRs as your working through it.

I added a couple of minor comments, but nothing blocking.

src/web-component.html
cocomarine: These changes were done by Max and I checked and assumed they are to do with linter.

src/components/Editor/Runners/PythonRunner/SkulptRunner/SkulptRunner.jsx
cocomarine: Ah thanks! Not intentional, just copied and pasted from example 🫠 Will update it to use the i18n library.

zetter-rpf: Still worth checking that it falls back as expected (assuming we want to show en errors to other locales?)

src/web-component.html
cocomarine: I will revert the linter changes and only leave what we need.

PR #1494, opened 2026-06-10
Remove unused files and dependencies with Knip

zetter-rpf:
This uses the [Knip tool](https://knip.dev/) to remove unused code and dependencies, similar to https://github.com/RaspberryPiFoundation/editor-standalone/pull/770.

It seems that most removals are due to:

+ Code that was copied when this project was split from editor-standalone that was never used
+ Packages that are no longer needed as dependencies as we don't directly use them (but they may be still installed as dependencies of dependencies)
+ Packages that were installed by default as part of the ejecting of 'create react app'.

I've tested the build locally from classroom and editor and all project types are working as expected.

This takes us from around 2540 dependencies (170 direct) to 2264 (132 direct), although doesn't change the build size that much as the dependencies were unused.


DNR500 (approved):
Looks good :-)

Issue #932, opened 2024-02-26
Fix Pyodide `.whl` file loading on staging and production


loiswells97:
The `.whl` files are not being loaded up correctly on staging or production, meaning that some imports that are available locally do not work, for example `sense_hat` and `turtle`

PR #1497, opened 2026-06-12
Fix feedback panel in scratch e2e tests

DNR500:
issue: [1489](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1489)

## Summary

Upgrade editor-ui to React 19 (matching editor-standalone) and bump related dependencies so the Scratch classroom flow works again after standalone’s React 19 upgrade.

## Why the Scratch e2e test was failing

The test failed at the final step: **teacher opens a student’s Scratch remix** from Student work. The Scratch iframe never loaded because **`editor-wc` crashed** with:

`Objects are not valid as a React child`

That only happens on this step because it is the first time **`isViewingStudentWork`** is true, which registers the **feedback sidebar plugin**.

## Feedback plugin + sidebar: the underlying issue

Classroom (`editor-standalone`) passes a sidebar plugin into the web component (`editor-ui`) via `setSidebarPlugins`. That plugin includes **React elements and components created by standalone’s React** (`FeedbackIcon`, `panel()`, `buttons()`), which `Sidebar` then renders inside **editor-ui’s React tree**.

Standalone and editor-ui are **separate bundles with separate React instances**. Passing elements across that boundary is unsupported.

This was **latent for years** but broke when standalone moved to **React 19** (June 9, PR #932) while editor-ui stayed on React 18 — React 19 elements are rejected by React 18’s reconciler.

**Why it’s brittle:** even with both on React 19, this remains fragile. Two bundled React copies can appear to work for simple JSX but can still break (hooks, context, portaled UI). A robust long-term fix would avoid passing React elements across the web component boundary (e.g. mount callbacks / string icon names only).

## Fix

Bump **editor-ui to React 19.2.7** to align with **editor-standalone**. Verified locally: teacher → student Scratch remix flow works again.

## Other dependency bumps

| Package | Change | Why |
|---------|--------|-----|
| **react / react-dom** | 18 → **19.2.7** | Match standalone; fixes cross-version plugin crash |
| **react-toastify** | 8 → **11.1.0** | v8 uses removed `react-dom` `render()` API; v11 supports React 18/19. Updated toast positions and icon usage in `Notifications.js`; removed v8 `ToastContainer.defaultProps` workaround |
| **@react-three/fiber** | 8 → **9.6.1** | R3F v8 pairs with React 18 only; v9 required for React 19 (Astro Pi 3D simulator) |
| **@react-three/drei** | 9 → **10.0.0** | Companion to R3F v9 (helpers used by Astro Pi `Simulator` / `FlightCase`) |
| **@react-three/test-renderer** | 8 → **9.0.0** | Test support for R3F v9 |

## Scratch React 18 vendor aliases

```json
"scratchReactVendor": "npm:react@18.3.1",
"scratchReactDomVendor": "npm:react-dom@18.3.1"
```

The main app uses **React 19**, but the **Scratch iframe** (`scratch.html`) loads **`scratch-gui` via `<script>` tags** and expects **React 18 UMD** globals (`vendor/react.production.min.js`). React 19 **no longer ships UMD builds**, so we cannot copy from the main `react` dependency. These aliases install React 18 solely so webpack can copy UMD files into `build/vendor/` for the Scratch iframe.


zetter-rpf (commented):
> Two bundled React copies can appear to work for simple JSX but can still break (hooks, context, portaled UI). A robust long-term fix would avoid passing React elements across the web component boundary (e.g. mount callbacks / string icon names only).

I'd be interested to know what the alternatives are that would work well across the web component boundary (or if this is a good reason not to use a web component).

zetter-rpf (approved):
The approach looks good. Have you checked this change with the Astro Pi functionality? I think there are e2e tests, but it would be good to make sure nothing is visually different from the major updates to the `react-three` library which I think only Astro Pi uses.

DNR500:
I checked AstroPi that looked ok to me.

I've updated another library and checked the two other libraries - even though their peer deps don't state support for React 19 they should work ok for now. But I have created two tickets to remove them

- https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1499
- https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1500 

Also added the investigate ticket to look at the general approach - I'll make sure that covers the concerns around the web component boundary (or if this is a good reason not to use a web component).

- https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1501

PR #1503, opened 2026-06-18
Replace react-toggle with DS Switch

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1499

### Changes
- Bumped Design System React to 2.9.2 to use Switch component and updated colour variable. 
- Replaced react-toggle Toggle with Switch and removed react-toggle dependency
- Increased height of colour and timer controls to match the minimum height applied to `.rpf-switch` for a WCAG AAA touch-target requirement
- Updated colours as per Max's design

### Screen recordings
- Before
https://github.com/user-attachments/assets/4a280a36-b5d9-4a12-b461-30af7b64e854

- After
https://github.com/user-attachments/assets/35745f4e-701e-4eaf-a8a3-2fe2eb7187fa

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
     "react-tabs": "^3.2.3",
     "react-timer-hook": "^3.0.5",
     "react-toastify": "^11.1.0",
-    "react-toggle": "^4.1.3",
     "redux": "^4.2.1",
     "redux-oidc": "^4.0.0-beta1",
     "scratchReactDomVendor": "npm:react-dom@18.3.1",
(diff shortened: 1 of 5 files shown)


DNR500 (approved):
Thanks for doing this  🙂

We have seen some cypress query's in [learner-experience-integration-tests](https://github.com/RaspberryPiFoundation/learner-experience-integration-tests) sometimes fail when we have updated the design system version before so it might be worth keeping an eye on what happens with those.

Might be worth just asking for a design check in before merging this too, I wonder if the grey thumb and track outline might risk looking like a disabled state but I would trust Max's opinion over mine on that

PR #1504, opened 2026-06-18
Add native queries and remove react-container-query

abcampo-iry:
## Summary
- Replace `react-container-query` with a small internal `ResizeObserver` hook for the editor container breakpoint.
- Keep existing native CSS container-query layout behavior unchanged.
- Preserve the current 720px responsive behavior for editor/output resizing:
  - desktop: side-by-side layout, right resize handle, `50%` width / `100%` height
  - stacked: column layout, bottom resize handle, `100%` width / `50%` height
- Remove the unused `containerQueries.js` helper and the `react-container-query` dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "react": "^19.2.7",
     "react-app-polyfill": "^2.0.0",
     "react-confirm-alert": "^2.8.0",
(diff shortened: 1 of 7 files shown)


DNR500 (approved):
Nice thanks for doing this!

PR #1519, opened 2026-07-02
Add save disabled attribute

cocomarine:
Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/636
The new attribute is to be used in editor-standalone: https://github.com/RaspberryPiFoundation/editor-standalone/pull/982


src/components/ProjectBar/ProjectBar.jsx
```diff
@@ -39,7 +40,7 @@ const ProjectBar = ({ nameEditable = true }) => {
             type="tertiary"
           />
         </div>
-        {!projectOwner && !readOnly && (
+        {!projectOwner && !readOnly && !saveDisabled && (
```
abcampo-iry, line 43: Quesion which i'm not sure.
this hides the project-bar Save button, but SaveButton is also rendered from other places such as the mobile Projects panel and Download panel. 

Should the saveDisabled guard live in `SaveButton` itself, or be applied to all `SaveButton` call sites,  we cover all possible entry points?

abcampo-iry (commented):
Small comment to understand better

src/components/ProjectBar/ProjectBar.jsx
cocomarine: Good point, Berto. Will look into this. 
(MobileProject doesn't seem to have `SaveButton` at all. Not sure if that's intentional. Will check with the team. )

cocomarine: Actually MobileProject has SaveButton in the sidebar that appears when Steps button at the bottom is clicked.

abcampo-iry (approved):
I think is good, worth maybe having a look if we are covering the autosave paths for code editor, but i think this might be enough. Thanks

PR #1524, opened 2026-07-07
Add licences link to Scratch projects

cocomarine:
Part of https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1553

### Notes
- The licence text are in English and not localised.
  - We could have translated summaries perhaps in the future. 
- Couldn't use Design System Modal as it required a lot of overriding.
- Used the existing GeneralModal for now.
  - Although it looks fine locally, it may not in staging and prod like other modals currently. 
     - The Close button may not render properly, but will have to see in staging if that's the case
  - Max wrote this issue on the broken modals: https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1561
  
 ### Screenshots
- For blocks project:
Screenshot 2026-07-10 at 08 57 03
Screenshot 2026-07-10 at 08 56 43

- For non blocks project, no Licences link in the info panel:
Screenshot 2026-07-10 at 10 09 32

### DIFF
--- a/public/translations/en.json
+++ b/public/translations/en.json
@@ -284,6 +284,7 @@
     "cookies": "Cookies",
     "accessibility": "Accessibility",
     "safeguarding": "Safeguarding",
+    "licenses": "Licenses",
     "charity": "Raspberry Pi Foundation - UK registered charity 1129409",
     "settingsMenu": {
       "heading": "Settings",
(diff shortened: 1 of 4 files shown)


zetter-rpf (approved):
Great 👍

PR #1537, opened 2026-07-13
Fix variable name clashes

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/499

- Turned out caching clearing was working fine, but there was variable name clash. 
- When printed out the variable `name` of the example code, it was assigned to `sys`, indicating it was getting mixed up with the global variable `name` used in the `runPythonAsync` . 

Screenshot 2026-07-16 at 11 14 41

- Updated the variables and now it shows the expected error.

Screenshot 2026-07-16 at 11 17 29

### DIFF
--- a/cypress/e2e/spec-wc-pyodide.cy.js
+++ b/cypress/e2e/spec-wc-pyodide.cy.js
@@ -253,6 +253,39 @@ print(text_out)
       "NameError: name 'math' is not defined",
     );
   });
+
+  it("clears a user-defined variable named 'name' between code runs", () => {
+    runCode("name = 'Jane'\nprint(name)");
+    getPythonConsoleOutput().should("contain", "Jane");
+    runCode("print(name)");
+    getErrorMessage().should(
+      "contain",
+      "NameError: name 'name' is not defined",
+    );
+  });
+
+  it("clears a user-defined variable named 'sys' between code runs", () => {
+    runCode("sys = 1\nprint(sys)");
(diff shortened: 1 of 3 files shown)


cypress/e2e/spec-wc-pyodide.cy.js
```diff
@@ -253,6 +253,39 @@ print(text_out)
...
+  });
+
+  it("clears a user-defined variable named 'sys' between code runs", () => {
+    runCode("sys = 1\nprint(sys)");
+    getPythonConsoleOutput().should("contain", "1");
+    runCode("print(sys)");
+    getErrorMessage().should("contain", "NameError: name 'sys' is not defined");
+  });
+
+  it("does not silently reuse a stale 'name' variable in an input loop across runs", () => {
```
zetter-rpf: Is this test case needed or is it covered by the tests above?

I ask as these are browser based tests that are expensive to run - each one can take up to 10 seconds making it easy to slow down our build. 

I would like for us to do work to see if we can make them run faster, but for now I want to make sure each one is giving value.

zetter-rpf (approved):
Nice one, the cause wasn't obvious to me so thanks for getting to the bottom of it.

cypress/e2e/spec-wc-pyodide.cy.js
cocomarine: Thanks, removed the redundant test.

PR #1538, opened 2026-07-14
docs: add docs relating to host integration from feedback panel investigation

DNR500:
Notes relating to issue: [1501](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1501)


DNR500:
Moving this back to draft so it can be used to update the documentation further

PR #1542, opened 2026-07-16
Fix project name save button

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1563

- Could not use native Design System Buttons as it affected the whole page styling. 
- Instead, kept the wrapped DS Button and overridden styling of `btn` class.
Screenshot 2026-07-16 at 16 38 50


src/assets/stylesheets/ProjectName.scss
```diff
@@ -46,14 +46,23 @@
 
 .project-name__button {
   margin: $space-0-5;
-  inline-size: 48px;
-  block-size: 48px;
+  inline-size: 48px !important;
+  block-size: 48px !important;
+  padding: 0 !important;
```
maxelkins, line 51: I assume this is required since the button style overrides mess up the fact it should be square?

maxelkins (approved):

src/assets/stylesheets/ProjectName.scss
cocomarine: yes, otherwise they become rectangle.

PR #1546, opened 2026-07-20
Scope button overrides to wrapper

maxelkins:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1604

## Summary

**The goal of this PR is that existing buttons and styles should be largely unchanged.**

The button style overrides used to apply to every Design System button, whether it went through our button wrapper or not. This PR scopes those overrides to the wrapper only, so native Design System buttons aren't affected by our editor-specific styling. This is the start of allowing us to have the native Design System buttons and the overriden ones live side by side while we migrate.

- A `buttonWrapper` class was added to the button wrapper component, and the overrides were rewritten so they only apply to buttons carrying that class, instead of every button on the page.
- This wrapper class will be removed when `DesignSystemButton` is removed.

Scoping the overrides this way uncovered a few buttons that were quietly relying on the old, unscoped styles. This PR also fixes those:

- The file names in the Files panel turned black and unreadable in dark mode.
- The "New file" button's icon positioning uses prop now instead of CSS.
- The "Go to Projects" button now correctly fills width with prop instead of CSS.
- The download button's icon positioning uses prop now instead of CSS
- The download button on the projects panel had lost its styling completely (it was missing some classes, and had been for a while - the old override was quietly covering for it).
- The download button's icon turned black on hover in dark mode, because the colour it was supposed to use was never actually set.

For that last one, instead of patching the missing colour, the download, upload and save icons (`download.svg`, `upload.svg`, `save.svg`) were changed to just follow the button's text colour automatically. The same was done for the file type icons in the Files panel (`file.svg`, `python_file.svg`, `html_file.svg`, `css_file.svg`, `csv_file.svg`) to fix the dark mode legibility issue above. This is a pattern worth using more widely across all icons, but for now it's only applied where a real problem was found, to keep this change small.

## Other small changes in this PR

Happy to pull these out but the styles in particular are somewhat important for this PR.

- Formatted `web-component.html` (the local preview page) - no logic changes.
  - I know this can make reviewing this file a little tricky, but its repeating causing hassle with auto formatting
  - I've purposely done the format in a separate commit so changes can be viewed after the format
  - I've also left comments pointing out key changes
- Disabled `use_editor_styles` on the local preview page. Those styles aren't used any more and were affecting how buttons looked during testing.

## What's not in this PR

A few related problems came up while doing this work, but are being tracked separately rather than fixed here:
- Native buttons still don't render correctly without a wrapper, because of a styling conflict elsewhere in the app - https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1605
- Some components still use old CSS classes that should be cleaned up once native buttons work properly - https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1606
- The Files panel list items have some styling problems of their own (wrong naming, unset colours) - https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1610


src/web-component.html
```diff
@@ -88,41 +87,48 @@
...
-        const initialProjectObject = initialProject ? JSON.parse(initialProject) : null;
+        const initialProjectObject = initialProject
+          ? JSON.parse(initialProject)
+          : null;
         const newWebComp = document.createElement("editor-wc");
-        
+
         newWebComp.setAttribute("with_projectbar", "true");
         newWebComp.setAttribute("with_sidebar", "true");
-        newWebComp.setAttribute("use_editor_styles", "true");
```
maxelkins, line 96: This is the key change on this file

src/web-component.html
```diff
@@ -88,41 +87,48 @@
...
+        newWebComp.setAttribute("editable_instructions", "true");
         newWebComp.setAttribute("friendly_errors_enabled", "true");
         // see "Offline support" section in README for more details
         newWebComp.setAttribute("offline_enabled", "true");
         newWebComp.setAttribute(
           "sidebar_options",
           JSON.stringify([
+            "projects",
```
maxelkins: Also added this in to preview more panels

jamiebenstead (approved):
LGTM

PR #1559, opened 2026-07-28
Add window.mod guard

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1649

Added guard the drag handler on window.mod being ready to prevent race condition.


jamiebenstead (approved):
LGTM

PR #1561, opened 2026-07-29
Handle non-standard p5/py5 exceptions

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1652

## Possible problem
- Skulpt exceptions raised in p5/py5 sketches were escaping to window.onerror, reported by Sentry as the opaque "Non-Error exception captured with keys: $d, args, traceback".
- The captured object is a Skulpt Python exception with an empty traceback (constructed in JS, e.g. p5's _friendlyError), thrown from an async p5 callback , so it never reached our error handling.

## Changes 
- Route preload/setup/draw (whole body, incl. the per-frame preamble) through Sk.uncaughtException, and stop the sketch (stopOnError) so a pre-canvas failure can't leave the draw loop running.
- Harden handleError against non-standard exceptions (missing tp$str, empty traceback, no constructor).
- Aslo fixed flaky tests in `spec-wc-resize.cy.js` by updating the helper file:
  -  Hardened `dragHandle` to re-query the handle for trigger("mousedown") and made `loadPythonStarterProject` wait on #project-identifier so it resolves against the replacement <editor-wc> rather than the one about to be destroyed.

> [!NOTE]  
> This is defensive hardening for the async-escape paths; the exact production trigger for EDITOR-UI-ABW was a private learner sketch and could not be reproduced. Behaviour is unchanged. 

_Generated with the help of Claude_

## Further investigation
- There are other similar errors (current error in the top row). 
- Need to investigate if they share the same route cause or are different problems.
Screenshot 2026-08-05 at 11 34 56


src/components/Editor/Runners/PythonRunner/SkulptRunner/SkulptRunner.jsx
```diff
@@ -412,27 +412,37 @@ const SkulptRunner = ({
...
+      // err is normally a Skulpt exception (tp$str/traceback), but native
+      // code called from Python (e.g. the p5/py5 shims) can throw plain JS
+      // errors instead, which don't have that shape - fall back gracefully
+      // rather than letting this handler itself throw.
+      const rawDescription =
+        typeof err.tp$str === "function" ? err.tp$str().v : undefined;
+      const errorDescription = (
+        rawDescription ||
+        err.message ||
+        t("editor.errors.generalError")
```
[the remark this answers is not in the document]
cocomarine: Removed.

src/components/Editor/Runners/PythonRunner/SkulptRunner/SkulptRunner.jsx
```diff
@@ -412,27 +412,37 @@ const SkulptRunner = ({
...
+        err.traceback?.length > 1 &&
         fileName === "./sense_hat.py" &&
         ["RuntimeError", "ValueError"].includes(errorType)
       ) {
         lineNumber = err.traceback[1].lineno;
         fileName = err.traceback[1].filename;
       }
 
-      fileName = fileName.replace(/^\.\//, "");
+      fileName = fileName ? fileName.replace(/^\.\//, "") : fileName;
```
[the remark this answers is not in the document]
cocomarine: Fixed by adding lineNumber check for appending `on line ...` part to the message.

zetter-rpf:
Do you know of any examples of p5 code that can cause this kind of error?  I'm trying to understand if the error we show to the user would be clear in this cases (and that context would be useful to have in the PR)

cocomarine:
> Do you know of any examples of p5 code that can cause this kind of error? I'm trying to understand if the error we show to the user would be clear in this cases (and that context would be useful to have in the PR)

Thanks, Chris. I will add some more context to the PR note. I think it needs more investigating (perhaps along with other similar errors).

cocomarine:
@copilot resolve the merge conflicts in this pull request

cypress/e2e/spec-wc-skulpt.cy.js
```diff
@@ -60,6 +60,29 @@ describe("Running the code with skulpt", () => {
...
+    // preload/setup run before any canvas exists, so VisualOutputPane can't
+    // stop the sketch. The shim must stop it, otherwise p5 treats the callback
+    // as successful and carries on into setup/draw - here draw() would print,
+    // which is asserted never to happen.
+    runCode(
+      "from p5 import *\n\ndef preload():\n\traise ValueError('boom in preload')\ndef setup():\n\tsize(400, 400)\ndef draw():\n\tprint('draw ran')\nrun(frame_rate=2)",
+    );
+
+    getErrorMessage().should("contain.text", "ValueError: boom in preload");
+    getSkulptRunner().should("not.contain", "draw ran");
```
DNR500, line 73: Nice that this checks draw never runs - that’s the important bit for the stuck loading indicator

public/shims/processing/p5/p5-shim.js
```diff
@@ -1510,47 +1510,76 @@ const $builtinmodule = function (name) {
...
+        try {
+          if (Sk.globals["settings"] && isPy5Version) {
+            Sk.misceval.callsimArray(Sk.globals["settings"]);
+          }
+          if (Sk.globals["setup"]) {
+            Sk.misceval.callsimArray(Sk.globals["setup"]);
+
+            for (const cb of Object.keys(callBacks)) {
+              if (Sk.globals[cb]) {
+                sketch[callBacks[cb]] = new Function(
```
DNR500, line 1549: Small note - not a blocker.

If preload, setup, or draw fails, we show the error and stop the sketch. If something like mouse_pressed fails, we still show the error but the sketch keeps running. Fine for this PR - might be worth matching later but I don't think it necessarily needs to be done on this PR

DNR500 (approved):
Looks good to me :-) 

Tried the preload/draw examples locally - was also nice to see that the Visual Output loading indicator actually stops too as part of this.

Image

cypress/helpers/editor.js
```diff
@@ -93,7 +93,11 @@ const dragHandle = (getHandle, { deltaX = 0, deltaY = 0 }) => {
     const clientX = left + width / 2;
     const clientY = top + height / 2;
 
-    cy.wrap($handle).trigger("mousedown", { button: 0, clientX, clientY });
+    // Re-query the handle instead of reusing $handle. If the app re-renders
+    // while trigger() is waiting for the element to become actionable, a
+    // wrapped element is detached for good and trigger() burns its full
+    // timeout; a query chain retries against the current element.
+    getHandle().trigger("mousedown", { button: 0, clientX, clientY });
```
DNR500, line 100: Nice fix - dropping cy.wrap($handle) and refinding the handle before mousedown is a good call.

The flake comes from a race - while we’re mid-drag (or even just waiting to start it), the app can rerender and replace the resize handle with a new DOM node. Cypress then keeps poking the old one until it times out. Requerying means we always hit the current handle, which should clear most of these failures. Waiting for #project-identifier before trusting the Run button is a good companion change for the same class of “element got swapped out underneath us” problem.

If this still flakes occasionally, we could harden a bit further without changing the app - e.g. wait until the handle’s position/size stops shifting before we drag, especially in the stacked layout test where things settle more slowly near the layout breakpoints. Happy to leave that until we see whether this is enough on its own

PR #1572, opened 2026-08-05
Upgrade to Node 24

zetter-rpf:
Node 20 reached end-of-life in April 2026 and we're seeing warnings now in GitHub Actions.

Also update setup node and remove some duplication of the node version


adrian-rpf (approved):
LGMT - nice use of .tool-versions

PR #1575, opened 2026-08-06
Do not set usedLeds unless a LED is on

### LINKED ISSUE #1577: `set_rotation(270, True)` erroneously marks `usedLEDs` as true
In Astro Pi Mission Zero the participants write a program that needs to fulfil a number of criteria, one of which is using the LEDs on the SenseHAT.

If you use the [staging site](https://staging-missions.astro-pi.org/mz/code_submissions/new), the following program ticks the "Used LEDs" checkbox when it shouldn't.

```python
# Import the libraries
from sense_hat import SenseHat
from time import sleep

# Set up the Sense HAT
sense = SenseHat()
# Rotates & **redraws** with `set_pixels`
sense.set_rotation(270, True)
# LEDs are not being used, the "Used LEDs" checkbox should not be checked
```

A suggestion would be to make sure in `set_pixel` the colour value is > 0, and in `set_pixels` that at least one pixel that has a non-zero value.

carlospena-irysan:
## Context

In Astro Pi Mission Zero the participants write a program that needs to fulfil a number of criteria, one of which is using the LEDs on the SenseHAT.

If you use the [staging site](https://staging-missions.astro-pi.org/mz/code_submissions/new), the following program ticks the "Used LEDs" checkbox when it shouldn't.

```python
# Import the libraries
from sense_hat import SenseHat
from time import sleep

# Set up the Sense HAT
sense = SenseHat()
# Rotates & **redraws** with `set_pixels`
sense.set_rotation(270, True)
# LEDs are not being used, the "Used LEDs" checkbox should not be checked
```

staging-missions astro-pi org_mz_code_submissions_new

Both the `setpixels` and the `setpixel` functions were emmiting the usedLeds = true message, even when the values passed were zero on every colour channel (id est, a led on "OFF").

In general that wouldn't be much of a problem as it doesn't make sense to set a pixel to zero when it was already off. But the bug was manifest when calling `sense.set_rotation(270, True)` as this function was internally getting the values of the board and redrawing it by calling `setpixels` again. 
So for a board with all leds in OFF, just rotating it, would end up emmitting the usedLeds = True message.

---

This PR makes sure we don't set usedLeds = True on setpixel or setpixels unless there's at least one non-zero value.

Closes #1577

### DIFF
--- a/cypress/e2e/missionZero-wc.cy.js
+++ b/cypress/e2e/missionZero-wc.cy.js
@@ -105,6 +105,15 @@ it("confirms LEDs used when display set", () => {
   getResults().should("contain", '"usedLEDs":true');
 });
 
+it("doesn't show LEDs used when display is all set to 0", () => {
+  runCode(
+    "from sense_hat import SenseHat\nsense = SenseHat()\nsense.set_pixels([[0,0,0]] * 64)",
+  );
+
+  cy.scrollTo("bottom");
+  getResults().should("contain", '"usedLEDs":false');
+});
+
 it("picks up calls to input()", () => {
   runCode("name = input('What is your name?')\nprint('Hello', name)");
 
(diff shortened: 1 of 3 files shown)


patch0 (approved):
LGTM! 🚀 🧑‍🚀

PR #1582, opened 2026-08-07
Editable paginated instructions

zetter-rpf:
Related to https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1681

This adds editable paginated markdown instructions

The editor already had support for paginated instructions for the project site, and editable instructions for classroom. This combines both.

https://github.com/user-attachments/assets/5b027602-09fe-4553-81b9-9e334a1b9b74

Note that the top (tabs) and bottom (prev/next buttons) of the panel isn't yet styled as designs. This will be done as separate issues

See commits for more.


src/components/Menus/Sidebar/InstructionsPanel/InstructionsPanel.jsx
```diff
@@ -64,49 +93,43 @@ const InstructionsPanel = () => {
...
+                fill={true}
+                textAlways={true}
+                small={true}
+              />,
+            ]
           : []
       }
       Footer={
-        hasMultipleSteps ? () => <ProgressBar panelRef={panelRef} /> : undefined
+        hasInstructions && (hasMultipleSteps || onEditTab)
```
DNR500, line 112: In the non-editable case, we still render the ProgressBar footer whenever we have instructions and the internal tabIndex happens to be 0 (I think that tabIndex defaults to 0 even though the tabs aren’t shown). I don’t know what the preferred behaviour is here but that can make the pagination UI appear when we probably only want it for the “more than 1 step” case. 

Maybe we should make the footer render condition match the design rule directly (e.g. only show this pagination UI when steps.length > 1), instead of depending on the edit tab state?

DNR500 (approved):
I think this is looking good.

I’ve left a comment that more asking about the the pagination visibility/behaviour.

Separately, I was wondering how we’re confident this works when older projects are still using the “step-less” instructions format. Do we expect both formats to be live at the same time (so we should test both)?

src/components/Menus/Sidebar/InstructionsPanel/InstructionsPanel.jsx
zetter-rpf: Good catch, I will look at this case

zetter-rpf:
> Separately, I was wondering how we’re confident this works when older projects are still using the “step-less” instructions format. Do we expect both formats to be live at the same time (so we should test both)?

Yes they will be both used for a time, and we could do a migration in editor-api to switch to one format after this is deployed.

The work I did to make sure we could handle both are in was shipped in https://github.com/RaspberryPiFoundation/editor-ui/pull/1580 by adding a new selector and there are some unit tests in there. I think there is coverage missing in the InstructionPanel component so I'll add another unit test there.

PR #1584, opened 2026-08-10
Add block text highlighting in instruction panel

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1690

## Block-type text highlighting
- Used [rpf-markdown-core](https://github.com/RaspberryPiFoundation/rpf-markdown-core) (built by the Code Club team) for rendering instructions, chosen for its inline Kramdown class support.
- Refactored the css file so that they apply to both light and dark mode without breaking others. 

| Before   | After |
| -------- | ------- |
| Screenshot 2026-08-10 at 17 02 22  | Screenshot 2026-08-10 at 17 04 13    |

## Prism and other fixes
- While doing this work, I found Prism had stopped working possibly due to the partial Vite migration. Prism provides syntax highlighting, line numbering and line highlighting in code blocks of instructions. Classroom and ExpCS instructions don't use these currently, but CCP instructions do (and CCP team also noticed this when running projects site locally along with latest editor-ui).
   Screenshot 2026-08-10 at 17 02 57

- The existing `vite.lib.js` configured `babel-plugin-prismjs` via @vitejs/plugin-react's babel option. But the plugin-react v6 no longer has, so the config was ignored. Replaced with explicit imports in src/utils/prism.js. 
- In dev environment, `vite-plugin-node-polyfills` prepends a shim banner to every pre-bundled dependency without terminating the last statement. Prism's plugin files start with (function(){…})(), so it was parsed as an argument to the shim and then short-circuited away, resulting in line-numbers and line-highlight not registered. Worked around with optimizeDeps.exclude in vite.config.js. 
- Jest couldn't load any test importing rpf-markdown-core. Its CJS build requires marked and scratchblocks, both ESM-only with no CommonJS entry. Updated transformIgnorePatterns to exempt both, matched against the full path so the nested rpf-markdown-core/node_modules/scratchblocks copy is also covered.

## Can our users use any other features that the rpf-markdown-core provides?
- Not yet. There are some styling issues with others which we need to address first. 
- CCP team is also open to collaboration if we require certain set of functionalities from or improvements to the library.

| Spacing issues   | issues with dark mode |
| -------- | ------- |
| Screenshot 2026-08-10 at 17 04 34 | Screenshot 2026-08-10 at 17 19 52 |

### DIFF
--- a/src/assets/stylesheets/Instructions.scss
+++ b/src/assets/stylesheets/Instructions.scss
@@ -8,48 +8,14 @@
 .project-instructions {
   block-size: 100%;
 
-  code.block3control {
-    background-color: var(--scratch-control-primary);
-  }
-  code.block3events {
-    background-color: var(--scratch-events-primary);
-  }
-  code.block3extensions {
-    background-color: var(--scratch-pen-primary);
-  }
-  code.block3looks {
-    background-color: var(--scratch-looks-primary);
-  }
-  code.block3motion {
-    background-color: var(--scratch-motion-primary);
-  }
-  code.block3myblocks {
-    background-color: var(--scratch-more-primary);
-  }
-  code.block3operators {
-    background-color: var(--scratch-operators-primary);
-  }
-  code.block3sensing {
-    background-color: var(--scratch-sensing-primary);
-  }
-  code.block3sound {
-    background-color: var(--scratch-sounds-primary);
-  }
-  code.block3variables {
-    background-color: var(--scratch-data-primary);
+  @each $block, $category in $scratch-block-categories {
+    code.block3#{$block} {
+      background-color: var(--scratch-#{$category}-primary);
+    }
   }
 
   @media (prefers-contrast: more) {
-    code.block3control,
-    code.block3events,
-    code.block3extensions,
-    code.block3looks,
-    code.block3motion,
-    code.block3myblocks,
-    code.block3operators,
-    code.block3sensing,
-    code.block3sound,
-    code.block3variables {
+    code[class*="block3"] {
       color: var(--rpf-black);
     }
   }
@@ -416,10 +382,10 @@
 }
 
 .--dark .project-instructions {
-  pre {
-    background-color: $rpf-grey-850;
-  }
-  code {
+  // Exclude scratchblocks references, otherwise this would tie on specificity
+  // with their category colours above and win by coming later in the file.
+  pre,
+  code:not([class*="block3"]) {
     background-color: $rpf-grey-850;
   }
 }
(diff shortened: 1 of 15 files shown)


yarn.lock
```diff
@@ -11926,6 +11977,21 @@ __metadata:
   languageName: node
   linkType: hard
 
+"happy-dom@npm:^20.9.0":
```
zetter-rpf, line 11980: I thought happy-dom was like JS dom and often used for test envs so surprised to see it here - is it an actual run time dependency of RPF markdown or just a dev one? Same with the ws (websocket) library below.

src/components/Menus/Sidebar/InstructionsPanel/InstructionsStep/InstructionsStep.jsx
```diff
@@ -1,20 +1,13 @@
...
-import Prism from "prismjs";
+import { processEditorProject } from "@raspberrypifoundation/rpf-markdown-core";
+import Prism from "../../../../../utils/prism";
 import { scratchblocksInit } from "../../../../../utils/scratchblocks";
 
-const markdownRenderer = new marked.Renderer();
-markdownRenderer.link = function (data) {
-  return `<a href="${data.href}" target="_blank" rel="noreferrer"
-    }">${data.text}</a>`;
-};
```
zetter-rpf, line 10: is this functionality in RPF markdown? If not, do we need it?

yarn.lock
zetter-rpf: (doesn't have to be fixed now as long as the bundle size isn't significantly bigger)

src/assets/stylesheets/Instructions.scss
```diff
@@ -8,48 +8,14 @@
...
-  }
-  code.block3sound {
-    background-color: var(--scratch-sounds-primary);
-  }
-  code.block3variables {
-    background-color: var(--scratch-data-primary);
+  @each $block, $category in $scratch-block-categories {
+    code.block3#{$block} {
+      background-color: var(--scratch-#{$category}-primary);
+    }
```
zetter-rpf, line 14: Max suggested also supporting "\`Looks\`{:block-type="looks"}" in https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1682, were you going to do that as part of this or in another change?

zetter-rpf (approved):
Nice one, I appreciate the time you've taken to understand RPF markdown and prism and to improve instructions as a whole.

Added a few comments, the only thing that might be a blocker to merging is the change in how links are rendered

maxelkins:
@cocomarine could we look at adding some more user friendly class names - happy to discuss this. Can also do in a follow up PR

yarn.lock
cocomarine: It is being used for testing but, yes, it is under `dependencies` rather than dev one. Will check with CCP team why that's the case.

src/assets/stylesheets/Instructions.scss
cocomarine: Yes, I will open another issue/PR to do this.

src/components/Menus/Sidebar/InstructionsPanel/InstructionsStep/InstructionsStep.jsx
cocomarine: Yes, we do need it. The equivalent override was added as `applyExternalLinkAttributes` in InstructionsStep.jsx.

PR #1585, opened 2026-08-10
Change demo page to navigate to project URLs

zetter-rpf:
We're now using this demo page more in our E2E tests. Clicking on the navigation link in the tests is a problem as it means the editor is initialized twice. This can slow down the tests and make the tests flakey if they are asserting an element that is removed from the page.

Doing this also enables you to refresh the page in development without loosing the project which is nice.

Thanks to @cocomarine and @DNR500 for discovering this might be causing some intermittent failures for the resize tests. This was fixed as part of #1561, this is a more general change to prevent similar issues for other tests in the future.


maxelkins:
> Doing this also enables you to refresh the page in development without loosing the project which is nice.

HUGE

cocomarine (approved):

PR #1586, opened 2026-08-11
Enable GitHub Advanced Security/CodeQL

mwtrew:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,99 @@
...
+      contents: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - language: javascript-typescript
+            build-mode: none
+          - language: python
+            build-mode: none
```
zetter-rpf, line 47: I'd be interested to know if it's worth checking python files or not - I'm not sure if there will be security risks from them since they run in the sandboxed python editor.

zetter-rpf (approved):

.github/workflows/codeql.yml
mwtrew: Yep, I'm curious to see what it turns up but might be worth disabling later

PR #1589, opened 2026-08-11
fix: clean up preview page

maxelkins:
## Tidy up the web component preview page

Cosmetic cleanup of the local demo page at `web-component.html` — no changes to the editor itself.

- Restructure the layout as a `main` CSS grid, and group the debug output (results, project identifier, run button) into its own panel
- Rewrite the styles: consistent `rem` spacing, logical properties, `box-sizing`
- Enable the `projects` sidebar panel and `editable_instructions` so both can be previewed locally
- Re-indent the inline script to standard Prettier formatting
- Open the file panel explicitly in the pyodide e2e specs, they relied on it being the sidebar's default panel, which `editable_instructions` changes

Best reviewed with whitespace changes hidden — most of the line count is the script re-indent.

Screenshot 2026-08-11 at 15 58 12

## Why 

- Makes the preview more responsive (particularly the picker and the debug menu)
- Prettier and cleaner
- Shows more panels

https://github.com/user-attachments/assets/6ed199c9-38bb-4225-a765-49a747f9e62d

### DIFF
--- a/cypress/helpers/editor.js
+++ b/cypress/helpers/editor.js
@@ -127,12 +127,10 @@ export const dragEditorResizeHandle = (
 export const dragSidebarResizeHandle = ({ deltaX = 80 } = {}) =>
   dragHandle(() => getSidebarResizeHandle(), { deltaX, deltaY: 0 });
 
-// The sidebar option toggles, so only click it when the panel is closed. Which
-// panel opens by default depends on the host's attributes, so tests that need
-// the file list should call this rather than assume it is already showing.
-export const openFilePanel = () =>
+export const ensureFilePanelOpen = () =>
   getFilePanelButton().then(($button) => {
-    if (!$button.hasClass("sidebar__bar-option--selected")) {
+    const alreadyOpen = $button.hasClass("sidebar__bar-option--selected");
+    if (!alreadyOpen) {
       cy.wrap($button).click();
     }
   });
--- a/cypress/helpers/editor.js
+++ b/cypress/helpers/editor.js
@@ -34,6 +34,9 @@ export const getFileButtonByName = (filename) =>
 const getSettingsButton = () =>
   getEditorShadow().find("[title='Settings']").first();
 
+const getFilePanelButton = () =>
+  getEditorShadow().find("[title='Project files']").first();
+
 export const getProgramInput = () =>
   getEditorShadow().findByRole("textbox", { name: "Text input" });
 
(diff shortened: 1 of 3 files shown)


web-component.html
```diff
@@ -75,84 +100,87 @@
...
-      document.addEventListener("editor-projectIdentifierChanged", (e) => {
-        document.getElementById("project-identifier").innerText = e.detail;
-      });
+        document.addEventListener("editor-runCompleted", (e) => {
+          document.getElementById("results").innerText = JSON.stringify(
+            e.detail,
+          );
+        });
 
-      const sampleBar = document.getElementById("sample-projects-bar");
```
maxelkins, line 99: dead code

cypress/helpers/editor.js
```diff
@@ -124,8 +127,15 @@ export const dragEditorResizeHandle = (
...
+// The sidebar option toggles, so only click it when the panel is closed. Which
+// panel opens by default depends on the host's attributes, so tests that need
+// the file list should call this rather than assume it is already showing.
 export const openFilePanel = () =>
-  getEditorShadow().find("[title='Project files']").click();
+  getFilePanelButton().then(($button) => {
+    if (!$button.hasClass("sidebar__bar-option--selected")) {
+      cy.wrap($button).click();
+    }
+  });
```
zetter-rpf, line 136: Comments like this can be an indication that the code itself isn't clear. While comments are useful sometimes, if we can make sure the code itself conveys what the comment is trying to say, we get to improve the code and we don't need to maintain a separate comment (which could get out of sync with the code).

If the code was the following, does the comment add anything?

```
export const ensureFilePanelOpen = () =>
  getFilePanelButton().then(($button) => {
    const alreadyOpen = $button.hasClass("sidebar__bar-option--selected")
    if (!alreadyOpen) {
      cy.wrap($button).click();
    }
  });
```

(or if `alreadyOpen || cy.wrap($button).click()` if preferred)

zetter-rpf (approved):
Nice improvement!

I don't actually know how that bottom bar is used, I think it's in some tests? It seems like some of might be redundant. Maybe one for another time.

Left a very optional point about comments in code.

maxelkins:
> I don't actually know how that bottom bar is used, I think it's in some tests? It seems like some of might be redundant. Maybe one for another time.

Yeah I believe it's used in some tests didn't want to remove it for now but agree something to look at.

Will improve that other comment, thanks!

PR #1594, opened 2026-08-12
Remove CCP heading styles

cocomarine:
Closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1708

## Summary
- Editor-ui web-component provides `host_styles` attribute and projects-ui already passes some styles using it, so I wanted to try this approach for moving the CCP heading styles to projects-ui. 
- Tried out this idea by pair-programming with CCP team's Ram and it worked!
- Ram will carry on with this and merge that in to projects-ui. 
- The CCP-specific styles are now removed from editor-ui and the headings are back to normal:
  Screenshot 2026-08-12 at 16 20 32

## Order of things
- We merge in this PR and release new editor-ui soon.
- Projects-ui is currently not on a latest version of editor-ui which still carries the CCP-specific styles so it won't break.
- CCP team is planning to bump editor-ui to the latest version next week, so as long as Ram's PR goes in before then, it should be fine.

### DIFF
--- a/src/assets/stylesheets/Instructions.scss
+++ b/src/assets/stylesheets/Instructions.scss
@@ -84,43 +84,6 @@
     }
   }
 
-  // Steps in CCP style
-  [id^="step-"] {
-    @include typography.style-1(bold);
-
-    background: #ecf8ee;
-    border-radius: $space-0-5;
-    color: #276d32;
-    display: inline-block;
-    margin: $space-1-5 0 0 0;
-    padding: $space-0-25 $space-0-5;
-  }
-
-  // "Now run your code" in CCP style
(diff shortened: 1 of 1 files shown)


zetter-rpf (approved):
Great!

src/assets/stylesheets/Instructions.scss
maxelkins, line 99: Is this still needed? `c-project-code` doesnt appear on rendered code blocks but does it on the projects site?

cocomarine: Looks like it, yes. As discussed, will make a list and work on them in a separate issue / PR so that I can go through them with CCP team.

PR #1600, opened 2026-08-17
Projects UI sample

rammodhvadia:
Adds a sample to the `/web-component.html` page for projects-ui
Also adds a themes folder and support for loading a css file for adding a custom theme to the sample project


web-component.html
```diff
@@ -7,7 +7,8 @@
     <title>Editor Web component</title>
     <script type="module" src="/src/web-component.jsx"></script>
     <style>
-      @import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100..900;1,100..900&display=swap");
+      /* Lexend is used by the projects-ui sample theme */
+      @import url("https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100..900;1,100..900&family=Lexend:wght@100..900&display=swap");
```
mwtrew, line 11: Will this mean an extra font gets loaded whenever the component loads, even outside of CCP? If so it might be worth exploring moving that import to the theme CSS if possible.

mwtrew (approved):
Looks good in general, one query below.

web-component.html
rammodhvadia: I did try to load this through the css theme, but didn't have much luck

PR #1604, opened 2026-08-19
refactor: bump design system and refactor theme styles

maxelkins:
> 🤖 Created with AI assistance
> 👨‍💻 Edited, reviewed and steered by @maxelkins

## Summary

Bumps `@raspberrypifoundation/design-system-react` to v2.10.3 (design-system-core v2.4.9), which moves Button's CSS custom properties onto `:root, :host` so they're reachable by inheritance. Simplifies our dark-theme button overrides in `InternalStyles.scss` to rely on that inheritance instead of redeclaring every token on `.rpf-button` itself.

## Changes

- Bump `@raspberrypifoundation/design-system-react` from `^2.10.2` to `^2.10.3`.
- Move the dark-theme Button default token overrides from `.rpf-button` up to `.--dark` directly, now that they cascade via inheritance.
- Keep the `:disabled` and `.rpf-button--secondary` overrides scoped to their own selectors, since those are conditional on state/variant rather than defaults.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@codemirror/view": "^6.3.0",
     "@hello-pangea/dnd": "^16.2.0",
     "@juggle/resize-observer": "^3.3.1",
-    "@raspberrypifoundation/design-system-react": "^2.10.2",
+    "@raspberrypifoundation/design-system-react": "^2.10.3",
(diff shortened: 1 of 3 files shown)


jamiebenstead (approved):
LGTM

PR #1607, opened 2026-08-21
Migrate more jest tests to vitest

zetter-rpf:
This follows on from #1570 and migrates more files to vitest

Most of the tests required renaming from `jest` to `vi`, and a few mocking changes.
Most of the changes to `setupTest.vitest.js` were imported from the jest `setupTests.js` file.

I have migrated the remaining files on a branch, I've just kept them separate to reduce the changeset as some of them are slightly more complex to migrate.

*changes made with Claude*

PR #1608, opened 2026-08-24
Migrate remaining jest tests to vitest

zetter-rpf:
This follows on from #1570 and https://github.com/RaspberryPiFoundation/editor-ui/pull/1607 and migrates the remaining tests to vitest.

+ These tests required setting up a more complex mock for the worker.
+ This also removes jest from the repo

See commits for more

As doing this I think there are a few things we should do after:
+ Install and configure babel vitest
+ Understand if we need babel-eslint and look at removing it
+ Fix any warnings coming from vitest

Changes made with Claude


DNR500 (approved):
Nice!

PR #1609, opened 2026-08-24
1430 image loading blocking p5 javascript projects

### LINKED ISSUE #1430: Image loading blocking p5 javascript projects
## Summary

The `draw-a-cat` project works locally, but does not load correctly in the Raspberry Pi code editor preview when the sketch uses `cat.png`.

Project:
[https://editor.raspberrypi.org/en/projects/draw-a-cat](https://editor.raspberrypi.org/en/projects/draw-a-cat)

## Actual behaviour

In the editor preview:
- the page shows `Waiting to load cat.png...`
- the sketch stays stuck on `Loading...`
- the image is not rendered

## Expected behaviour

The project should load `cat.png` and render the sketch in the editor preview, the same way it does locally.

## Evidence

Attached screenshots show:
- editor preview stuck loading
- local version loading successfully and rendering the cat image
Image
Image

## Reproduction

1. Open the `draw-a-cat` project in the editor
2. Run the preview
3. Observe that the image does not load and the sketch does not complete startup

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/editor-ui/issues/1430


src/components/Editor/Runners/HtmlRunner/HtmlRenderer.jsx
```diff
@@ -17,15 +17,25 @@ import {
...
+        `['"]${escapeForRegExp(mediaFile.filename)}['"]`,
+        "g",
+      ); // prevent substring matches
       updatedProjectFile.content = updatedProjectFile.content.replaceAll(
         find,
-        replace,
+        () => `"${mediaFile.url}"`, // callback, so $-sequences in the URL are not expanded
       );
     });
   }
```
zetter-rpf: I was thinking about ways to simplify this, I since `escapeForRegExp` looks hard to understand and might be easy to get wrong. 

One way might be avoid the need to build a regex by calling `replaceAll` with the different strings:

```js
updatedProjectFile.content.replaceAll(
        `"${mediaFile.filename"}"`,
        replace,
        () => `"${mediaFile.url}"`, // callback, so $-sequences in the URL are not expanded
      ).replaceAll(
        `'${mediaFile.filename"}'`,
        replace,
        () => `"${mediaFile.url}"`, // callback, so $-sequences in the URL are not expanded
     )
```

Slightly less efficient, but avoids needing to construct a regular expression.

If you don't think that would work, we could keep the regexp, but it would be good to have a least another unit test that covers the need to escape it.

zetter-rpf (approved):
Nice one of fixing this bug, I think your the third person to look at it and the first to make progress 😀 

I've added a suggestion on how to simplify the code by avoiding the need for escaping regexes.

jamiebenstead:
> I was thinking about ways to simplify this, I since `escapeForRegExp` looks hard to understand and might be easy to get wrong.
> 
> One way might be avoid the need to build a regex by calling `replaceAll` with the different strings:
> 
> ```js
> updatedProjectFile.content.replaceAll(
>         `"${mediaFile.filename"}"`,
>         replace,
>         () => `"${mediaFile.url}"`, // callback, so $-sequences in the URL are not expanded
>       ).replaceAll(
>         `'${mediaFile.filename"}'`,
>         replace,
>         () => `"${mediaFile.url}"`, // callback, so $-sequences in the URL are not expanded
>      )
> ```
> 
> Slightly less efficient, but avoids needing to construct a regular expression.
> 
> If you don't think that would work, we could keep the regexp, but it would be good to have a least another unit test that covers the need to escape it.

@zetter-rpf Thanks the review, made a change to use `replaceAll`, what do you think?

zetter-rpf:
> @zetter-rpf Thanks the review, made a change to use replaceAll, what do you think?

Nice one, I think that's a lot simpler

PR #1611, opened 2026-08-26
Improve instruction headings

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/420
With these changes we get [1750](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1750) for free

Updated heading sizes:
image


DNR500 (approved):
Nice, might be worth just checking if any other header sizes h4–h6 that we might want to support as follow up

PR #1613, opened 2026-08-26
Only show scroll bars for students instructions when needed

DNR500:
## Summary
issue: [1707](https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1707)

Short project instructions no longer show a scrollbar. The bar still appears when the text is too long to fit.

Screenshot of instructions without scroll bar

### DIFF
--- a/src/assets/stylesheets/Instructions.scss
+++ b/src/assets/stylesheets/Instructions.scss
@@ -6,7 +6,10 @@
 @use "./scratch_colours" as *;
 
 .project-instructions {
-  block-size: 100%;
+  // Student preview sizes to its content so short steps are not scrollable.
+  &:has(.c-instruction-tabs) {
+    block-size: 100%;
+  }
 
   @each $block, $category in $scratch-block-categories {
     code.block3#{$block} {


src/assets/stylesheets/Instructions.scss
```diff
@@ -6,7 +6,10 @@
 @use "./scratch_colours" as *;
 
 .project-instructions {
-  block-size: 100%;
+  // Student preview sizes to its content so short steps are not scrollable.
+  &:has(.c-instruction-tabs) {
+    block-size: 100%;
+  }
```
maxelkins, line 12: Looks good.

I assume in the react component we already know when this tab is rendered - we could use a modifer class instead for slightly better browser support. Although `:has()` is baseline now 🤷

maxelkins (approved):

jamiebenstead (approved):

src/assets/stylesheets/Instructions.scss
DNR500: `:has()` has been in place for a while now so I think it safe to use

PR #1615, opened 2026-08-28
Sanitise project instructions before rendering them

jamiebenstead:
Instruction step HTML was assigned to innerHTML unsanitised, so a step could run script in the host page's origin, where the login session is stored. Every step now goes through DOMPurify, covering both the `instructions` attribute and steps loaded with a project.

Two tags DOMPurify drops by default are kept because real content needs them: `use`, which draws the icons in a scratchblocks SVG, and `iframe`, which is then restricted to the editor's own origins so the embedded project viewers in project instructions keep working.

closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1738

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,18 @@ Styles from the parent application can be passed to the web component in a few d
 }
 ```
 
+#### Instructions Sanitisation
+
+Instruction steps are rendered into the page with `innerHTML`, so every step is
+sanitised with DOMPurify first (see `src/utils/sanitiseInstructions.js`). This
+applies to steps passed in the `instructions` attribute and to steps loaded with
+a project. Scripts, event handler attributes, `javascript:` and `data:` URLs and
+stylesheets outside a scratchblocks SVG are removed.
+
+`<iframe>` is kept only when its `src` is `https://editor.raspberrypi.org` or
+`https://staging-editor.raspberrypi.org`, so that the embedded project viewers
+used by project site content keep working. Any other frame is removed.
+
 #### Instructions Styling
 
 Classes from the stringified HTML passed to the web component in the `instructions` attribute are being used to style the project steps in the instructions panel.
--- a/README.md
+++ b/README.md
@@ -251,12 +251,14 @@ Styles from the parent application can be passed to the web component in a few d
 Instruction steps are rendered into the page with `innerHTML`, so every step is
 sanitised with DOMPurify first (see `src/utils/sanitiseInstructions.js`). This
 applies to steps passed in the `instructions` attribute and to steps loaded with
-a project. Scripts, event handler attributes, `javascript:` and `data:` URLs and
-stylesheets outside a scratchblocks SVG are removed.
+a project.
 
-`<iframe>` is kept only when its `src` is `https://editor.raspberrypi.org` or
-`https://staging-editor.raspberrypi.org`, so that the embedded project viewers
-used by project site content keep working. Any other frame is removed.
+Scripts, event handler attributes and `javascript:` URLs are removed, as are
+stylesheets outside an SVG and any stylesheet that loads CSS from elsewhere.
+`data:` URLs are still allowed on images. `<iframe>` is kept only when it points
+at the `editor.raspberrypi.org` or `staging-editor.raspberrypi.org` domains, so
+that the embedded project viewers in project site content keep working; any
+other frame is removed.
 
 #### Instructions Styling
 
--- a/src/utils/sanitiseInstructions.js
+++ b/src/utils/sanitiseInstructions.js
@@ -1,0 +1,52 @@
+import DOMPurify from "dompurify";
+
+// Some project steps embed the editor's own project viewer to show a worked
+// example. Frames from anywhere else are removed
+const EMBED_ORIGINS = [
+  "https://editor.raspberrypi.org",
+  "https://staging-editor.raspberrypi.org",
+];
+
+const isProjectViewer = (src) => {
+  try {
+    return EMBED_ORIGINS.includes(new URL(src, window.location.href).origin);
+  } catch {
+    return false;
+  }
+};
+
+const sanitiseConfig = {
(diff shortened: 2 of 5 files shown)


src/utils/sanitiseInstructions.js
```diff
@@ -0,0 +1,52 @@
...
+
+const remove = (node) => node.parentNode?.removeChild(node);
+
+purifier.addHook("uponSanitizeElement", (node, { tagName }) => {
+  if (tagName === "iframe" && !isProjectViewer(node.getAttribute("src"))) {
+    return remove(node);
+  }
+
+  // A stylesheet anywhere else would restyle the rest of the editor
+  if (tagName === "style" && !node.closest("svg")) {
```
DNR500: This keeps any `<style>` that’s inside an `<svg>`, not just the ones that come from Scratch blocks. 

That’s a problem because that CSS isn’t limited to the picture — it can restyle the rest of the editor. Someone who can put HTML in the instructions could hide buttons, overlay a fake panel, or load CSS from somewhere else. It’s not the same as running script on the logged-in page, but they can still mess with the UI.

A few ways we could try to tighten this:

1. Still allow `<style>` in SVGs, but drop `@import` and external `url(...)`. Keep `url(#something)` so Scratch filters still work. This is the lightest change.
2. Only keep `<style>` on SVGs that actually look like Scratch output.
3. Strip all `<style>` tags and put the Scratch stylesheet back ourselves after sanitising.

I’d go with 1 unless Scratch already needs external `url(` in those styles. Myabe try something like this

```javascript
if (tagName === "style") {
  const svg = node.closest("svg");
  const css = node.textContent ?? "";
  const loadsOutsideCss =
    /@import/i.test(css) ||
    /url\s*\(\s*['"]?(https?:|\/\/)/i.test(css);
  if (!svg || loadsOutsideCss) {
    return remove(node);
  }
}
```

src/utils/sanitiseInstructions.js
```diff
@@ -0,0 +1,52 @@
...
+    return remove(node);
+  }
+
+  // Same document references only, so a `use` cannot pull in outside markup
+  if (tagName === "use" && !node.getAttribute("href")?.startsWith("#")) {
```
DNR500: The sanitizer only checks `href` on `<use>` — that’s what we use for icons in Scratch blocks, like the green flag. There are two ways to point at the icon: `href` (the one we check) and `xlink:href` (the older one, which we ignore).

So this gets through:
```html
<svg><use href="#ok" xlink:href="https://somewhere-else/icon.svg"></use></svg>
```
`href="#ok"` looks local, so we keep the tag. The browser may still fetch the file from `xlink:href`.

If someone only sets `xlink:href` to an outside URL, we already remove it. This is the mixed case: the “safe” attribute is local, the old one is not.
Unlikely to run script on the logged-in page. It can still make the browser request a file from a site we don’t control, which doesn’t match “same document only”.

Fix: require both to start with `#`, or drop `xlink:href` so only `href` counts. Something like:

```javascript
const isLocalRef = (value) => value == null || value.startsWith("#");
if (
  tagName === "use" &&
  !(
    isLocalRef(node.getAttribute("href")) &&
    isLocalRef(node.getAttribute("xlink:href"))
  )
) {
  return remove(node);
}
```

src/utils/sanitiseInstructions.test.js
```diff
@@ -0,0 +1,210 @@
...
+    expect(iframe.getAttribute("src")).toEqual(src);
+    expect(iframe.getAttribute("width")).toEqual("600");
+    expect(iframe.getAttribute("allowfullscreen")).not.toBeNull();
+  });
+
+  test.each([
+    "https://evil.example/x",
+    "https://editor.raspberrypi.org.evil.example/x",
+    "//evil.example/x",
+    "/en/embed/viewer/x",
```
DNR500, line 115: I think we could try adding a test that checks a relative address when the page itself is already on `editor.raspberrypi.org`.

That way we check relative iframe `src` values while the page is already on one of the allowed domains, not only on the fake test page. 

I would say this is a low priority as the sanitiser already rejects these, this just makes sure a test would fail if that behaviour slipped.

Might look something like..

```javascript
test("Removes relative embeds even when the page is on an allowed origin", () => {
  jsdom.reconfigure({ url: "https://editor.raspberrypi.org/en/projects/foo" });
  expect(parse(embed("/en/embed/viewer/x")).querySelector("iframe")).toBeNull();
  expect(
    parse(embed("//editor.raspberrypi.org/en/embed/viewer/x")).querySelector(
      "iframe",
    ),
  ).toBeNull();
});
```

README.md
```diff
@@ -246,6 +246,18 @@ Styles from the parent application can be passed to the web component in a few d
 }
 ```
 
+#### Instructions Sanitisation
+
+Instruction steps are rendered into the page with `innerHTML`, so every step is
+sanitised with DOMPurify first (see `src/utils/sanitiseInstructions.js`). This
+applies to steps passed in the `instructions` attribute and to steps loaded with
+a project. Scripts, event handler attributes, `javascript:` and `data:` URLs and
```
DNR500: This overstates what the sanitizer does.

`javascript:` URLs are stripped, but `data:` is still allowed on images (including `data:image/svg+xml,…`). That isn’t really script-on-the-page, but we shouldn’t tell integrators we remove all `data:` URLs if we don’t.

Same for the iframe lines just below: they read as if `src` has to be exactly those two hosts. We actually keep any page on those sites. Given we’re happy with that, this could just say we keep frames from those two sites and remove the rest.

This is low priority really, docs only, not a hole in the XSS fix. Something like:

> Scripts, event handler attributes and `javascript:` URLs are removed. Stylesheets outside an SVG are removed. `<iframe>` is kept only when it points at the `editor.raspberrypi.org` or `staging-editor.raspberrypi.org` domains; any other frame is removed.

DNR500 (approved):
Nice! This is looking like it should do the job.

I've added a couple of points where things might be tighten up a little more - that maybe be worth a look at before merging

PR #1621, opened 2026-09-04
Release from git tags instead of package.json

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1757

- gates the latest_version file on the tag being on main, replacing the old check that matched the tag against the package version
- requires lint, test and cypress to pass before a tag deploys


.github/workflows/ci-cd.yml
```diff
@@ -134,6 +134,7 @@ jobs:
 
   deploy-tag:
     if: github.ref_type == 'tag'
+    needs: [lint, test, test-cypress]
```
zetter-rpf, line 137: Nice idea

zetter-rpf (approved):
Nice one! 

Do you think it's worth doing a release now to test it?

PR #1635, opened 2026-09-10
Emit host-facing events for stop, save, download and sidebar

patch0:
These interactions were tracked by calling `window.plausible` directly from components. Custom events are the documented channel out of the shadow DOM, and the host owns both the vendor choice and the context the payload needs (projectSlug, locale, consent), so emit events rather than pushing to window.dataLayer from here.

```js
  editor-runStopped         { embedded }
  editor-saveTriggered      { loggedIn }
  editor-projectDownloaded  { projectType }
  editor-sidebarToggled     { panel, expanded }
```


src/components/SaveButton/SaveButton.test.jsx
```diff
@@ -9,14 +9,19 @@ import useIsOnline from "../../hooks/useIsOnline";
...
     document.addEventListener("editor-logIn", logInHandler);
+    document.addEventListener("editor-saveTriggered", (e) =>
+      saveTriggeredHandler(e.detail),
+    );
   });
 
   beforeEach(() => {
     useIsOnline.mockReturnValue(true);
+    saveTriggeredHandler.mockClear();
   });
```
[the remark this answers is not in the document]
patch0: Already resolved

src/components/Menus/Sidebar/Sidebar.jsx
```diff
@@ -215,9 +216,18 @@ const Sidebar = ({
     }
   }, [dispatch, nextDefaultOption, optionIsAvailable, selectedSidebarOption]);
 
+  // Emit here rather than from each control: on mobile toggleOption never
+  // collapses, so a control computing `expanded` itself would report
+  // closures that don't happen
   const updateOption = (nextOption) => {
     setOption(nextOption);
     dispatch(setSidebarOption(nextOption));
+    document.dispatchEvent(
```
adrian-rpf, line 225: **`editor-sidebarToggled` is skipped when the effect reassigns the panel**

`updateOption` is the only emitter, but the effect at lines 210-217 also changes the selected panel and bypasses it by calling `setOption` + `dispatch(setSidebarOption(...))` directly.

Two concrete cases where the host's view goes stale:

- The user has the **images** panel open and deletes the last image. `removeOption("images", projectImages)` drops it from `menuOptions`, `optionIsAvailable` goes false, and the effect switches to `nextDefaultOption`. No event, so the host still thinks `images` is expanded.
- On first load with an `autoOpen` plugin, `nextDefaultOption` changes while `selectedSidebarOption === undefined`, so the effect opens the plugin panel — again with no event. The host never learns which panel is initially open.

README.md:163 promises the event fires "When the selected sidebar panel changes", so either route those transitions through a shared emitter or narrow the doc wording to user-initiated toggles.

src/components/RunButton/StopButton.jsx
```diff
@@ -16,6 +17,7 @@ const StopButton = ({ embedded = false, className, ...props }) => {
     if (window.plausible) {
       window.plausible(`Stop button${embedded ? " embedded" : ""}`);
     }
+    document.dispatchEvent(runStoppedEvent({ embedded }));
```
adrian-rpf, line 20: **`editor-runStopped` fires when no code run is being stopped, and never fires for Scratch**

The dispatch sits above the `if (codeRunTriggered)` guard, so it also covers the draw-only path.

Concrete case: a p5/py5 sketch imports `./p5/__init__.js`, so `SkulptRunner.jsx:268` dispatches `triggerDraw()`. When the Python program finishes, `codeRunTriggered` goes false and `editor-runCompleted` has already fired, but `drawTriggered` stays true so `RunnerControls.jsx:19` keeps rendering this button. Clicking stop now dispatches only `stopDraw()` — yet the host sees `runStarted → runCompleted → runStopped` for a run that already ended.

The mirror problem: `ScratchContainer.jsx:89` emits `editor-runStarted` for Scratch, but no Scratch path emits `editor-runStopped`, so Scratch starts are permanently unmatched. Worth deciding whether this event means "a code run was stopped" (then guard it on `codeRunTriggered`) or "the user pressed stop" (then say so in the README, next to `runStarted`/`runCompleted`, which currently read as a lifecycle triple).

src/components/SaveButton/SaveButton.jsx
```diff
@@ -36,9 +39,10 @@ const SaveButton = ({ className, type, fill = false }) => {
     if (window.plausible) {
       window.plausible("Save button");
     }
+    document.dispatchEvent(saveTriggeredEvent({ loggedIn: !!user }));
     document.dispatchEvent(logInEvent);
```
adrian-rpf, line 43: **`logInEvent` is still dispatched when the user is already logged in**

`SaveButton` returns `null` only when `projectOwner`, so a **logged-in non-owner** sees the Save button. Clicking it now emits `editor-saveTriggered {loggedIn: true}` and then, on this line, `editor-logIn` — two contradictory signals in the same tick. A host that wires `editor-logIn` to its sign-in redirect will bounce an already-authenticated user mid-save.

This predates the PR, but the guard is now a one-liner because `!!user` is computed directly above:

```suggestion
    if (!user) {
      document.dispatchEvent(logInEvent);
    }
```

src/components/ProjectBar/ScratchProjectBar.jsx
```diff
@@ -37,6 +38,13 @@ const ScratchProjectBar = ({ nameEditable = true }) => {
     enabled: canSave,
   });
 
+  // Scratch saves here rather than through SaveButton, so the event has to be
+  // emitted from both or Scratch projects report no saves at all
+  const onClickSave = () => {
+    document.dispatchEvent(saveTriggeredEvent({ loggedIn: !!user }));
```
adrian-rpf, line 44: **`loggedIn` can only ever be `true` here**

`onClickSave` is only reachable from the button rendered under `showSaveButton` (line 53), which requires `canSave`, which is `Boolean(user && !readOnly && !previewMode)` (line 35). So the Scratch path always reports `loggedIn: true`.

Two consequences: a host segmenting save intent by `loggedIn` gets a permanently skewed split versus the `SaveButton` path, and the new test at `ScratchProjectBar.test.jsx:152` would pass just as well if this were a hard-coded `true` — it cannot catch a regression.

Either drop the field on this path, or move the dispatch above the `canSave` gate so logged-out save intent on Scratch is actually reported (which is what the `SaveButton` path does via its `header.loginToSave` variant).

src/components/RunButton/StopButton.jsx
```diff
@@ -16,6 +17,7 @@ const StopButton = ({ embedded = false, className, ...props }) => {
     if (window.plausible) {
       window.plausible(`Stop button${embedded ? " embedded" : ""}`);
```
adrian-rpf, line 18: **The `window.plausible` calls are kept alongside the new events**

The PR description says these interactions "were tracked by calling `window.plausible` directly from components" and that custom events are the right channel — but every plausible call survives: here (17-19), `DownloadButton.jsx:38-40`, `SaveButton.jsx:39-41`. Once a host wires the new events into its own analytics, each stop/save/download is counted twice.

The sidebar is the messier case: `SidebarBar.jsx:55-58` and `:63-65` still fire `"Expand file pane"` / `"Collapse file pane"`, and `SidebarBarOption.jsx:10-12` fires `"Side menu open project files"` for the `file` panel only. Those now overlap `editor-sidebarToggled` while reporting a panel-blind shape, so the two channels will disagree on what the sidebar did.

If the removal is deliberately a follow-up, worth saying so in the README so hosts know which channel is authoritative in the meantime.

src/events/WebComponentCustomEvents.js
```diff
@@ -29,9 +29,21 @@ export const runCompletedEvent = (detail) =>
 export const runStartedEvent = (detail) =>
   webComponentCustomEvent("editor-runStarted", detail);
 
+export const runStoppedEvent = (detail) =>
```
adrian-rpf, line 32: **`runStopped`'s payload does not match its lifecycle siblings**

`WebComponentProject.jsx:135-141` dispatches `runStartedEvent({ step, projectIdentifier, projectType })`, and `runCompletedEvent` carries the run payload. `runStoppedEvent` carries only `{ embedded }`.

A host building a run funnel keyed on `projectIdentifier` has nothing to join the stop onto and has to fall back to "most recent run", which breaks as soon as the project changes mid-session. `embedded` is also the one field the host already knows — it chose the embed mode — while `projectIdentifier` / `projectType` are the ones it cannot derive from inside the shadow DOM.

src/components/DownloadButton/DownloadButton.jsx
```diff
@@ -37,6 +38,9 @@ const DownloadButton = (props) => {
     if (window.plausible) {
       window.plausible("Download");
     }
+    document.dispatchEvent(
```
adrian-rpf, line 41: **The event is emitted before the download is attempted**

README.md:157 documents this as "When the user downloads the project", but it fires at the top of the handler, before any work.

For Scratch the very next call is `postMessageToScratchIframe`, whose `getScratchIframeContentWindow` (`src/utils/scratchIframe.js:1-5`) dereferences `document.querySelector("editor-wc").shadowRoot.querySelector("iframe[title='Scratch']").contentWindow` with no guards — if the iframe is not mounted this throws and no file is produced, but the event has already gone out. For python/html, `zip.generateAsync` or `FileSaver.saveAs` can reject after the fact too.

Either move the dispatch after the work succeeds, or reword the README the way `editor-saveTriggered` does ("fires on the click, not on a successful save").

Minor, same line: `project.project_type` is read unguarded while line 48 uses `project?.name` for the same object — worth making consistent.

src/components/Menus/Sidebar/Sidebar.test.jsx
```diff
@@ -93,6 +108,25 @@ describe("When project has images", () => {
       expect(screen.queryByText("filePanel.files")).not.toBeInTheDocument();
     });
 
+    test("Clicking collapse dispatches editor-sidebarToggled as collapsed", () => {
```
adrian-rpf, line 111: **The mobile case the design hinges on is untested**

The comment at `Sidebar.jsx:219-221` justifies emitting from `updateOption` precisely because "on mobile `toggleOption` never collapses, so a control computing `expanded` itself would report closures that don't happen". Both new tests run in the default (desktop) viewport, so only the `!isMobile` branch is exercised.

If someone later moves the dispatch into `toggleOption` or `SidebarBarOption` — the exact change the comment warns against — this suite stays green while mobile users emit a spurious `{ expanded: false }` on every re-tap of the open panel. Mocking `useMediaQuery` to match `MOBILE_MEDIA_QUERY` and asserting `sidebarToggledHandler` is *not* called when the active option is tapped again would pin the invariant down.

src/components/DownloadButton/DownloadButton.test.jsx
```diff
@@ -276,6 +291,13 @@ describe("When project is Scratch", () => {
     expect(JSZip).not.toHaveBeenCalled();
     expect(FileSaver.saveAs).not.toHaveBeenCalled();
   });
+
+  test("editor-projectDownloaded reports the project type", () => {
```
adrian-rpf, line 304: **Only the Scratch path is covered**

`onClickDownload` returns early for `code_editor_scratch` at `DownloadButton.jsx:45`, so this test never reaches the zip branch — which is the common case for python and html projects.

If the detail key were renamed, or `project.project_type` swapped for something only Scratch projects carry, non-Scratch downloads would report the wrong `projectType` and this suite would still pass. The `Downloading project with name set` describe already renders a non-Scratch project and clicks the button, so one extra assertion there closes the gap.

src/components/ProjectBar/ScratchProjectBar.test.jsx
```diff
@@ -20,6 +20,21 @@ vi.mock("react-router-dom", async () => ({
...
+beforeAll(() => {
+  document.addEventListener("editor-saveTriggered", onSaveTriggered);
+});
+
+afterAll(() => {
+  document.removeEventListener("editor-saveTriggered", onSaveTriggered);
+});
+
+beforeEach(() => {
+  saveTriggeredHandler.mockClear();
```
adrian-rpf: **Redundant with the `vi.clearAllMocks()` below**

A second top-level `beforeEach` at line 117 already calls `vi.clearAllMocks()`, which clears `saveTriggeredHandler` too. Today the ordering works out (this hook is registered first, `clearAllMocks` second), but a reader now has to check both hooks to know what state each test starts in, and the two are only correct by registration order. Lines 34-36 can go.

src/components/DownloadButton/DownloadButton.test.jsx
```diff
@@ -15,6 +15,21 @@ vi.mock("jszip-utils", () => ({
...
+beforeAll(() => {
+  document.addEventListener("editor-projectDownloaded", onProjectDownloaded);
+});
+
+afterAll(() => {
+  document.removeEventListener("editor-projectDownloaded", onProjectDownloaded);
+});
+
+beforeEach(() => {
+  projectDownloadedHandler.mockClear();
```
adrian-rpf, line 30: **The listener stub is copy-pasted into five test files**

The same six lines — `vi.fn()`, `e => handler(e.detail)`, `beforeAll` add, `afterAll` remove, `beforeEach` clear — appear in `DownloadButton.test.jsx:18-31`, `Sidebar.test.jsx:27-40`, `ScratchProjectBar.test.jsx:23-36`, `StopButton.test.jsx:8-22` and `SaveButton.test.jsx:11-18`, differing only in the event name. They have already drifted: `SaveButton` registers inside a `describe` while the others register at file scope, and `ScratchProjectBar` duplicates the clear.

The repo already keeps shared test helpers next to the code (`src/utils/renderWithProviders.jsx`). A `spyOnCustomEvent("editor-projectDownloaded")` helper that returns the spy and wires up add/remove/clear would drop ~50 lines and make the registration scope uniform.

src/components/SaveButton/SaveButton.jsx
patch0: If this is an issue, we should bump it to another PR

src/components/RunButton/StopButton.jsx
patch0: Yup I didn't want to remove plausible events.  Maybe the PR description is a bit enthusiastic.

src/components/RunButton/StopButton.jsx
patch0: This mirrors the current behaviour for the plausible thing.  I'm trying to keep this PR minimal.

src/components/DownloadButton/DownloadButton.jsx
patch0: Mirrors the plausible event.

adrian-rpf (approved):
A few comments but I think you can merge when ready.

PR #1645, opened 2026-09-18
Fix local Scratch asset fixtures so the teapot project loads

jamiebenstead:
Scratch requests assets at /internalapi/asset/<md5>.<ext>/get/ with a trailing slash. Vite treats that as a directory request and 404s, so the local fixtures never loaded and sprites rendered as broken images.

Add a dev/preview middleware that serves the fixture file at the same path without the trailing slash, and set the content type from the md5ext — the fixture file is named `get` with no extension, so Vite sent it empty.

Also add the Stage backdrop fixture (cd21514d...svg), which was never committed. Copied from scratch-gui's own default-project.

No production impact: the middleware only defines configureServer and configurePreviewServer, so none of it is emitted into the build. The new fixture does ship, since public/ is copied to build/, but that matches the teapot fixture already tracked there. Nothing reads these paths in production — Scratch takes its asset host from REACT_APP_API_ENDPOINT, which points at editor-api on a different origin, and nginx serves no /api route.

image


zetter-rpf (approved):
Thanks for improving this,

It's useful the asset loads locally now. Although could still be a little confusing if it doesn't work for branch previews.

DNR500 (approved):

PR #1646, opened 2026-09-18
Use Scratch Foundation's scratch-gui instead of our fork

jamiebenstead:
closes https://github.com/RaspberryPiFoundation/digital-editor-issues/issues/1772

Upstream v15 added GUIStorage.getLibraryAssetUrl, which does what our fork's libraryAssetUrlTemplate prop did, so the fork no longer carries any code we need. Depend on @scratch/scratch-gui 15.1.1 from public npm and supply getLibraryAssetUrl from scratch-frame instead.

Also serve the Scratch chunks from both /chunks/ and /scratch-gui/chunks/. Workers such as scratch-storage's fetch-worker resolve against the server root, while lazy-loaded chunks such as v15's new paper-source resolve against the bundle's publicPath. 13.7.3 had no paper-source, so one base was enough; with v15 a missing fetch-worker leaves project loading hanging with no error at all.

Leaves .yarnrc.yml and NPM_AUTH_TOKEN in place. Removing the GitHub Packages config is deliberately a follow-up, so reverting to the fork stays easy until this is proven in production.


DNR500 (approved):
Nice one. Looks good to me. I checked the actual new Scratch package and the pieces it needs (the asset URL setting, the file-serving change) all line up correctly, so this should work as intended.

One small thing: the setup instructions in the README still talk about the old private package and won't match what's actually needed anymore. Not urgent and doesn't affect how the app runs - just needs updating at some point so it doesn't confuse anyone setting up the project fresh. Maybe this can be included on the clean up ticket. 

Approving.

